PR #34, opened 2026-04-08
Icon directory creation fails on Ubuntu due to AppArmor restriction

david-bartlett:
# Fix: Icon directory creation fails on Ubuntu due to AppArmor restriction

## Problem

On Ubuntu 26.04, the icon directory creation in step 13 fails with `Permission Denied`:

```
mkdir: Permission denied
  ✗ Webcam Fix (Book 5) exited with code 1
```

The failing line is:
```bash
sudo -u "$_relay_user" mkdir -p "$ICON_DEST"
```

Ubuntu's AppArmor policy blocks `sudo -u <user> mkdir` for paths under `/home` when
invoked from within a root (pkexec) context, even when the target user and path are
valid. This is a known Ubuntu 26.04 AppArmor restriction.

## Fix

Replace the `sudo -u` mkdir call with a plain `sudo mkdir` followed by a `chown` to
correct ownership:

```bash
# Before
sudo -u "$_relay_user" mkdir -p "$ICON_DEST"

# After
sudo mkdir -p "$ICON_DEST"
sudo chown -R "$_relay_user":"$_relay_user" "${_relay_home}/.local/share/icons"
```

The `chown` is applied from `icons` downward since `.local/share` already exists
on a clean Ubuntu install and we only want to fix ownership of directories we created.

## Testing

- Confirmed fix resolves the `Permission Denied` error on Ubuntu 26.04 (kernel 7.0,
  GNOME, AppArmor enabled)
- Confirmed no regression on Fedora 44 (KDE Plasma, SELinux)

### DIFF
--- a/webcam-fix-book5/install.sh
+++ b/webcam-fix-book5/install.sh
@@ -836,7 +836,8 @@ if [[ -d "$RELAY_DIR" ]]; then
 
     if [[ -n "$_relay_user" ]]; then
         ICON_DEST="${_relay_home}/.local/share/icons/hicolor/symbolic/apps"
-        sudo -u "$_relay_user" mkdir -p "$ICON_DEST"
+        sudo mkdir -p "$ICON_DEST"
+        sudo chown -R "$_relay_user":"$_relay_user" "${_relay_home}/.local/share/icons"
         for icon in camera-disabled-symbolic camera-switch-symbolic camera-video-symbolic; do
             if [[ -f "$RELAY_DIR/yaru-icons/${icon}.svg" ]]; then
                 sudo -u "$_relay_user" cp "$RELAY_DIR/yaru-icons/${icon}.svg" "$ICON_DEST/"
--- a/webcam-fix-libcamera/install.sh
+++ b/webcam-fix-libcamera/install.sh
@@ -1187,7 +1187,8 @@ if [[ -d "$RELAY_DIR" ]]; then
 
     if [[ -n "$_relay_user" ]]; then
         ICON_DEST="${_relay_home}/.local/share/icons/hicolor/symbolic/apps"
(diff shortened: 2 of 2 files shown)


Andycodeman:
Thanks for the fix and the thorough writeup, David!

One question before we merge: the `sudo -u "$_relay_user" cp` calls right after the mkdir fix (lines copying the SVG icons into `$ICON_DEST`) — did those succeed on your Ubuntu 26.04 test?

If AppArmor blocked `sudo -u mkdir` under `/home` in the pkexec context, it seems likely it would also block `sudo -u cp` to the same path. If the `cp` calls did work, great — just want to confirm. If not (or if they failed silently), we should convert those to the same pattern:

```bash
sudo cp "$RELAY_DIR/yaru-icons/${icon}.svg" "$ICON_DEST/"
sudo chown "$_relay_user":"$_relay_user" "$ICON_DEST/${icon}.svg"
```

Can you double-check that the icons actually landed in `~/.local/share/icons/hicolor/symbolic/apps/` with correct ownership after a fresh install?

david-bartlett:
@Andycodeman 

Good point. I'll run some tests.

david-bartlett:
@Andycodeman 

It appears to work fine on Ubuntu 26.04.

When the ~/.local/share/icons/hicolor is created (if needed) by root we dont have an apparmor issue.

The whole ~/.local/share/icons folder is then recursively chowned with $_relay_user.

Andycodeman (approved):
Looks good — thanks for the fix and for confirming the cp calls work. Merging!

Andycodeman:
Thanks David! Appreciate the fix and the testing on both Ubuntu and Fedora. Merging now.

Issue #35, opened 2026-04-09
Installing webcam-fix-book5 on Ubuntu 26.10 which includes IPU7 in the kernel


david-bartlett:
# Ubuntu 26.04 Webcam Fix — Changes & Notes

Tested on Samsung Galaxy Book 5 Pro (940XHA), Ubuntu 26.04 (Resolute Raccoon), kernel 7.0.0-12-generic, libcamera 0.7.0-1ubuntu2 (repo-installed).

@Andycodeman  I will include these minor changes as a PR once the previous one goes through.

---

## ⚠️ Do NOT install these packages on Ubuntu 26.04

The kernel already includes in-tree IPU7 support. Installing these overrides it with an incompatible DKMS version and breaks the camera:

- `intel-ipu7-dkms`
- `linux-modules-ipu7-generic`
- `linux-modules-ipu7-generic-hwe-24.04`
- `linux-main-modules-ipu7-7.0.0-12-generic`

`intel-vision-dkms` is safe and can be installed.

---

## Required packages

```
gstreamer1.0-libcamera
intel-vision-dkms
libcamera-ipa
libcamera-tools
libcamera-v4l2
libcamera0.7
libspa-0.2-libcamera
pipewire-libcamera
python3-libcamera
python3-picamera2
python3-v4l2
v4l-utils
v4l2loopback-dkms
v4l2loopback-utils
```

---

## User permissions

The user must be in the `video` group or WirePlumber cannot open `/dev/media0`:

```bash
sudo usermod -aG video $USER
```

Logout/login required after.

---

### `webcam-fix-book5/install.sh`

**IPA path detection** — add Ubuntu multiarch path:

```diff
     elif [[ -d "/usr/local/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
         IPA_PATH="/usr/local/lib/x86_64-linux-gnu/libcamera/ipa"
+    elif [[ -d "/usr/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
+        IPA_PATH="/usr/lib/x86_64-linux-gnu/libcamera/ipa"
     else
         IPA_PATH="/usr/lib/libcamera/ipa"
     fi
```

**SPA plugin detection** — add `/usr/lib` to search:

```diff
-    SPA_PLUGIN=$(find /usr/local/lib -path "*/spa-*/libcamera/libspa-libcamera.so" 2>/dev/null | head -1)
[...]

PR #36, opened 2026-04-09
Ubuntu 26.04 Compatibility Fixes

david-bartlett:
# PR: Ubuntu 26.04 Compatibility Fixes

Fixes Issue ([#35](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/issues/35#issue-4232727641))
## Summary

This PR fixes three issues that prevent the webcam fix from working on Ubuntu 26.04 (Resolute Raccoon) with repo-installed libcamera 0.7.0. The scripts were originally written and tested against Ubuntu 25.10 where libcamera had to be compiled from source into `/usr/local/`. On Ubuntu 26.04, libcamera 0.7.0 is available from the standard repos and installs to the standard Ubuntu multiarch path `/usr/lib/x86_64-linux-gnu/`.

## Test Environment

- **Hardware:** Samsung Galaxy Book 5 Pro (940XHA), Intel Lunar Lake
- **OS:** Ubuntu 26.04 (Resolute Raccoon), kernel 7.0.0-12-generic
- **libcamera:** 0.7.0-1ubuntu2 (repo-installed)

---

### 1. `webcam-fix-book5/install.sh` — IPA path detection

**Problem:** The Ubuntu IPA path detection block was missing `/usr/lib/x86_64-linux-gnu/libcamera/ipa` — the path where Ubuntu 26.04's repo-installed libcamera puts its IPA modules. Without this entry, the path fell through to the incorrect `/usr/lib/libcamera/ipa` fallback, causing libcamera to fail to find its IPA modules.

```diff
     elif [[ -d "/usr/local/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
         IPA_PATH="/usr/local/lib/x86_64-linux-gnu/libcamera/ipa"
+    elif [[ -d "/usr/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
+        IPA_PATH="/usr/lib/x86_64-linux-gnu/libcamera/ipa"
     else
         IPA_PATH="/usr/lib/libcamera/ipa"
     fi
```

### 2. `webcam-fix-book5/install.sh` — SPA plugin detection

**Problem:** The PipeWire SPA plugin search only looked in `/usr/local/lib`, missing the repo-installed plugin at `/usr/lib/x86_64-linux-gnu/spa-0.2/libcamera/libspa-libcamera.so`.

```diff
-    SPA_PLUGIN=$(find /usr/local/lib -path "*/spa-*/libcamera/libspa-libcamera.so" 2>/dev/null | head -1)
+    SPA_PLUGIN=$(find /usr/local/lib /usr/lib -path "*/spa-*/libcamera/libspa-libcamera.so" 2>/dev/null | head -1)
```

---

### 3. `webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh` — Full install for Ubuntu/Debian

**Problem:** The bayer fix script has two install modes — a full install (`ninja install`) which replaces the library and IPA modules together with matching signatures, and a so-only install which replaces only the `.so` files while preserving the distro's IPA `.sign` files.

On Ubuntu 26.04 the so-only install causes a signature mismatch. The patched library has a different build hash from the original, so the distro `.sign` files are rejected and libcamera falls back to `/usr/src/ipa`:

```
INFO IPAManager: libcamera is not installed. Adding '/usr/src/ipa' to the IPA search path
```

This results in no camera being detected despite the library being installed correctly. The fix is to use the full install path for Ubuntu/Debian, the same as already done for Arch Linux:

```diff
 elif [[ "$DISTRO" == "arch" ]]; then
     USE_FULL_INSTALL=true
     info "Full install: Arch Linux (so-only install breaks IPA path resolution)."
+elif [[ "$DISTRO" == "debian" || "$DISTRO" == "ubuntu" ]]; then
+    # Ubuntu/Debian: IPA modules must match the patched library's build hash.
+    # The .so-only approach preserves distro IPA .sign files which were built
+    # against the original library — causing signature mismatch and camera failure.
+    USE_FULL_INSTALL=true
+    info "Full install: Ubuntu/Debian (IPA modules must match patched library)."
 fi
```

---

### 4. `camera-relay` — IPA path detection

**Problem:** The `detect_ipa_path()` function was missing the `/usr/lib/` Ubuntu multiarch paths for both x86_64 and aarch64. These are added after the `/usr/local/`
entries so source builds still take priority over repo-installed libcamera.

```diff
     for path in /usr/local/lib/x86_64-linux-gnu/libcamera \
                 /usr/local/lib/aarch64-linux-gnu/libcamera \
+                /usr/lib/x86_64-linux-gnu/libcamera \
+                /usr/lib/aarch64-linux-gnu/libcamera \
                 /usr/lib64/libcamera /usr/lib/libcamera /usr/local/lib/libcamera \
                 /lib64/libcamera /lib/libcamera; do
```

### 5. `video` group check (all distros)

WirePlumber needs access to `/dev/media0` to expose the camera to PipeWire.
This device is owned by group `video` and the user must be a member or the
camera will not appear in any apps despite the kernel modules loading correctly.
This applies to all distros — some desktop installers add the user to `video`
automatically but it is not guaranteed.

The check is placed after the camera-relay installation block where
`_relay_user` is already set — this gives a reliable username regardless of
whether the script is run as root or as the logged-in user. Adding a user to
`video` when they are already a member is harmless.

```bash
if [[ -n "$_relay_user" ]]; then
    if ! groups "$_relay_user" | grep -q "\bvideo\b"; then
        echo "  Adding $_relay_user to video group..."
        sudo usermod -aG video "$_relay_user"
        echo "  ⚠ $_relay_user must log out and back in for the video group to take effect."
    else
        echo "  ✓ $_relay_user already in video group"
    fi
fi
```

---

### DIFF
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -259,6 +259,8 @@ detect_ipa_path() {
     local path
     for path in /usr/local/lib/x86_64-linux-gnu/libcamera \
                 /usr/local/lib/aarch64-linux-gnu/libcamera \
+                /usr/lib/x86_64-linux-gnu/libcamera \
+                /usr/lib/aarch64-linux-gnu/libcamera \
                 /usr/lib64/libcamera /usr/lib/libcamera /usr/local/lib/libcamera \
                 /lib64/libcamera /lib/libcamera; do
         if find "$path" -name "*.so" -path "*/ipa/*" 2>/dev/null | grep -q .; then
--- a/webcam-fix-book5/install.sh
+++ b/webcam-fix-book5/install.sh
@@ -572,6 +572,8 @@ elif [[ "$DISTRO" == "ubuntu" ]]; then
         IPA_PATH="/usr/local/lib/libcamera/ipa"
     elif [[ -d "/usr/local/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
         IPA_PATH="/usr/local/lib/x86_64-linux-gnu/libcamera/ipa"
+    elif [[ -d "/usr/lib/x86_64-linux-gnu/libcamera/ipa" ]]; then
+        IPA_PATH="/usr/lib/x86_64-linux-gnu/libcamera/ipa"
     else
         IPA_PATH="/usr/lib/libcamera/ipa"
     fi
@@ -586,7 +588,7 @@ fi
 # to /usr/local which PipeWire's systemd service doesn't search by default.
 SPA_PATH=""
 if [[ "$DISTRO" == "ubuntu" ]]; then
-    SPA_PLUGIN=$(find /usr/local/lib -path "*/spa-*/libcamera/libspa-libcamera.so" 2>/dev/null | head -1)
+    SPA_PLUGIN=$(find /usr/local/lib /usr/lib -path "*/spa-*/libcamera/libspa-libcamera.so" 2>/dev/null | head -1)
     if [[ -n "$SPA_PLUGIN" ]]; then
         # Extract the spa-0.2 directory (parent of libcamera/)
         SPA_PATH=$(dirname "$(dirname "$SPA_PLUGIN")")
@@ -856,6 +858,19 @@ if [[ -d "$RELAY_DIR" ]]; then
     fi
 else
     echo "  ⚠ camera-relay directory not found — skipping relay tool installation"
+fi
+
+# Ubuntu: ensure the user is in the video group so WirePlumber can access
+# /dev/media0. This must be done after _relay_user is set so we have a
+# reliable username regardless of whether the script is run as root.
+if [[ -n "$_relay_user" ]]; then
+    if ! groups "$_relay_user" | grep -q "\bvideo\b"; then
+        echo "  Adding $_relay_user to video group (required for /dev/media0 access)..."
+        sudo usermod -aG video "$_relay_user"
+        echo "  ⚠ $_relay_user must log out and back in for the video group to take effect."
+    else
+        echo "  ✓ $_relay_user already in video group"
+    fi
 fi
 
 # ──────────────────────────────────────────────
--- a/webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh
+++ b/webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh
@@ -868,6 +868,12 @@ elif [[ "$DISTRO" == "arch" ]]; then
     # so this is safe.
     USE_FULL_INSTALL=true
     info "Full install: Arch Linux (so-only install breaks IPA path resolution)."
+elif [[ "$DISTRO" == "debian" || "$DISTRO" == "ubuntu" ]]; then
+    # Ubuntu/Debian: IPA modules must match the patched library's build hash.
+    # The .so-only approach preserves distro IPA .sign files which were built
+    # against the original library — causing signature mismatch and camera failure.
+    USE_FULL_INSTALL=true
+    info "Full install: Ubuntu/Debian (IPA modules must match patched library)."
 fi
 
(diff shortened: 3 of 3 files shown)


Andycodeman:
Thanks David — this all looks good. One question before we merge:

The diff includes a change in `build-patched-libcamera.sh` (lines 933-936) that appends `/ipa` to `$LIBCAMERA_IPA_DIR` when setting `LIBCAMERA_IPA_MODULE_PATH`:

```diff
-        echo "export LIBCAMERA_IPA_MODULE_PATH=$LIBCAMERA_IPA_DIR" > "$IPA_ENV_FILE"
+        echo "export LIBCAMERA_IPA_MODULE_PATH=$LIBCAMERA_IPA_DIR/ipa" > "$IPA_ENV_FILE"
```

This isn't mentioned in the PR description. It looks correct — `LIBCAMERA_IPA_DIR` resolves to something like `/usr/lib/x86_64-linux-gnu/libcamera` (the parent), so the env var was pointing one level too high. But this is in the so-only install path, which after your Ubuntu/Debian full-install change only applies to Fedora.

Was this intentional? Did you hit this during testing or was it a separate bug you noticed?

david-bartlett:
@Andycodeman 

Sorry, I didn't release it wasn't in the description. 

I think because the file name was /etc/profile.d/libcamera-ipa-path.sh I assumed  it should have pointed to the ipa path and not the parent folder. 

---

We also appear to have a /etc/profile.d/libcamera-ipa.sh that points to the same path so i'm not sure why we need both.

I will double check tomorrow and figure out why we have both /etc/profile.d files and if the changes to /etc/profile.d/libcamera-ipa-path.sh are correct, or if the file is needed.

david-bartlett:
@Andycodeman 

Hi Andy, further investigation shows that Fedora always takes the SRPM path which sets `USE_FULL_INSTALL=true`, so the so-only install path has never actually been used on Fedora. 

Combined with our Ubuntu/Debian full install change, the so-only path is now effectively unreachable on all supported distros.

The only change I would make to build-patched-libcamera.sh would be:

```
-    IPA_ENV_FILE="/etc/profile.d/libcamera-ipa-path.sh"
+    IPA_ENV_FILE="/etc/profile.d/libcamera-ipa.sh"
```

---

If we keep the so-only code for future use, do you want me to submit a new PR that removes:

```
-        echo "export LIBCAMERA_IPA_MODULE_PATH=$LIBCAMERA_IPA_DIR" > "$IPA_ENV_FILE"
+        echo "export LIBCAMERA_IPA_MODULE_PATH=$LIBCAMERA_IPA_DIR/ipa" > "$IPA_ENV_FILE"
```
And adds:

```
-    IPA_ENV_FILE="/etc/profile.d/libcamera-ipa-path.sh"
+    IPA_ENV_FILE="/etc/profile.d/libcamera-ipa.sh"
```

Andycodeman:
@david-bartlett Good investigation — makes sense that the so-only path is effectively dead code now. No need for a separate cleanup PR, we can address it if that path ever becomes relevant again.

Merging as-is. Thanks for the thorough work on this — Ubuntu 26.04 repo-installed libcamera support is a big deal for users who don't want to compile from source. Appreciate the testing on both Ubuntu and Fedora!

Andycodeman (approved):
LGTM — merging.

david-bartlett:
@Andycodeman 

Sounds good.

I have another PR for webcam-fix-book5/install.sh which I will submit later for your consideration. 

It effectively installs all packages required for libcamera, v4l2, qcam etc for a given distribution. 

If it can't find a package it skips. 

This way I could get the camera fully working on a clean install of Ubuntu 26.04 and Arch (and Fedora) with minimal effort.

---

I thought it may end up with less support requests with people forgetting to install packages.

Issue #59, opened 2026-06-07
Hardware Test Report: Samsung Galaxy Book5 Pro 16 (960XHA) on NixOS


ang3lo-azevedo:
**Exact model:** Samsung Galaxy Book5 Pro 16" (960XHA)
**Distro and kernel version:** NixOS 26.11, Kernel 7.0.10-cachyos-lto
**Whether apps (Firefox, Zoom, etc.) can see the camera:** Yes, it is working with both PipeWire (native) and V4L2 (camera-relay) sources. The V4L2 source successfully rotates the image right-side up with `RELAY_COLOR_FILTER=videoflip method=vertical-flip`, and the colors are correct with the `bayer-fix-v0.7.patch`. There is a slight purple tint in some lighting, likely due to LSC mismatch with the flipped sensor, but the main color channels are decoding correctly. 

**Output of `cam -l`:**
```
[20:14:36.733790165] [395479]  INFO IPAManager ipa_manager.cpp:147 libcamera is not installed. Adding '/nix/store/src/ipa' to the IPA search path
[20:14:36.733874612] [395479]  INFO Camera camera_manager.cpp:340 libcamera v0.7.0
[20:14:36.746879054] [395482] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 15-0010': Unable to get rectangle 2 on pad 0/0: Inappropriate ioctl for device
[20:14:36.746908598] [395482]  WARN CameraSensor camera_sensor_legacy.cpp:402 'ov02e10 15-0010': The PixelArraySize property has been defaulted to 1928x1088
[20:14:36.746915861] [395482] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 15-0010': Unable to get rectangle 1 on pad 0/0: Inappropriate ioctl for device
[20:14:36.746959457] [395482]  WARN CameraSensor camera_sensor_legacy.cpp:413 'ov02e10 15-0010': The PixelArrayActiveAreas property has been defaulted to (0, 0)/1928x1088
[20:14:36.746963481] [395482] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 15-0010': Unable to get rectangle 0 on pad 0/0: Inappropriate ioctl for device
[20:14:36.746965832] [395482]  WARN CameraSensor camera_sensor_legacy.cpp:421 'ov02e10 15-0010': Failed to retrieve the sensor crop rectangle
[...]

Issue #49, opened 2026-05-10
v4l2loopback


noopduck:
Fedora doesn't package something called v4l2loopback

bat webcam-fix-book5/install.sh
760 │     # Install v4l2loopback if not present
 761 │     if ! modinfo v4l2loopback &>/dev/null 2>&1; then
 762 │         echo "  Installing v4l2loopback..."
 763 │         if [[ "$DISTRO" == "fedora" ]]; then
 764 │             sudo dnf install -y v4l2loopback 2>/dev/null || true

Andycodeman:
Thanks for the report @noopduck — you're right, Fedora has no `v4l2loopback` package. It lives in **RPM Fusion (free)** as `akmod-v4l2loopback`, and the failure was being silenced by `2>/dev/null || true`, so the relay would just be quietly broken.

Fixed in **[v0.3.42](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/releases/tag/v0.3.42)**. The Fedora branch of `webcam-fix-book5/install.sh` now:

1. Auto-enables the `rpmfusion-free` repo if it's missing (via the release RPM matched to your Fedora version).
2. Installs `akmod-v4l2loopback` + matching `kernel-devel`.
3. Runs `akmods --force` + `depmod -a` so the module is usable without a reboot.
4. Prints clear copy-paste recovery commands if the module still isn't available.

Could you pull the latest and re-run `webcam-fix-book5/install.sh`? If `dnf` asks you to accept the RPM Fusion GPG key, say yes. After the install, please confirm:

```bash
modinfo v4l2loopback | head -3
lsmod | grep v4l2loopback
ls /dev/video*
```

If anything still fails, paste the install.sh output and I'll dig in further. Going to leave this open until you confirm it works on your machine.

hywol:
dkms-v4l2loopback works well on Fedora. but, akmod didn't work on Fedora.

Tested on Galaxy Book4 Pro 125H + 16G RAM + 512G SSD South Korea domestic model.

Andycodeman:
Thanks @hywol — that's useful data, but I want to dig in before changing the default, because `akmod-v4l2loopback` is the standard Fedora path for v4l2loopback and is what RPM Fusion documents, so "didn't work" usually means one of a few specific things rather than akmod itself being broken.

Could you share a bit more so I can fix the right thing?

[...]

PR #60, opened 2026-06-07
feat(book5): NixOS support, Bayer fix v0.7, and V4L2 monitor improvements

ang3lo-azevedo:
This PR introduces several fixes and improvements for the Samsung Galaxy Book5 webcam, specifically targeting NixOS compatibility, new kernel behaviors, and stability for V4L2 clients like OBS.

### Bayer Fix v0.7 (Fixes "Broken pipe" on newer kernels)
Newer IPU7 drivers rigorously enforce V4L2 link validation. The previous `bayer-fix` altered the media bus `videoFormat` when `videoFlip` was enabled, causing `ioctl` link validation failures (`Inappropriate ioctl for device`) and resulting in a "Broken pipe".
- **Fix:** v0.7 of the patch leaves the `videoFormat` (kernel-facing) untouched to satisfy link validation. Instead, it recomputes the flipped Bayer layout and passes it exclusively to the `SoftwareISP`. The ISP now correctly decodes the flipped Bayer grid without angering the kernel driver.

### NixOS Module Fixes
- Added `RELAY_COLOR_FILTER = "videoflip method=vertical-flip"` to the `camera-relay` service when `videoFlip` is enabled. Previously, only `LIBCAMERA_FORCE_OV02E10_ROTATION` was being passed (which fixed the colors), but the geometric flip was omitted, leaving the V4L2 output physically upside down.

### Camera Relay Monitor Improvements
- Updated `camera-relay-monitor.c` to internally poll with a 100ms timeout.
- During the ~2-3s pipeline startup phase, the monitor now actively pumps black frames to the device. This prevents strict clients (like OBS, which has a 166ms `select()` timeout) from timing out and disconnecting before the first real frame arrives.

### Documentation
- Added hardware testing confirmation for Samsung Galaxy Book5 Pro 16" (960XHA) on NixOS with CachyOS kernel 7.0.10.
- Updated `README.md` and `webcam-fix-book5/README.md` to reflect the NixOS testing and community credits.

**Tested on:**
- Samsung Galaxy Book5 Pro 16" (960XHA)
- NixOS 26.11, Kernel 7.0.10-cachyos-lto
- Working perfectly with both OBS (V4L2) and PipeWire (native).


Andycodeman (changes requested):
Thanks for this — really thorough work, and most of it is exactly the right approach. The Bayer v0.7 patch (keep the kernel-facing `videoFormat` matching the bus code so link validation passes, and feed the true flipped Bayer order to the SoftwareISP via a separate `ispFormat`) is the correct layer to fix the newer-IPU7-kernel EPIPE/"broken pipe" failure, and the NixOS module generalization (`relayColorFilter`, `GST_PLUGIN_SYSTEM_PATH_1_0`, the loopback label handling) all look good.

Two things I'd like changed before merging, plus one follow-up:

**1. Please drop the two `ov02c10.yaml` CCM edits** (both `webcam-fix-book5/ov02c10.yaml` and `webcam-fix-libcamera/ov02c10.yaml`). Your test device (960XHA) is an **OV02E10** sensor, and this PR doesn't touch `ov02e10.yaml` — so the new colour matrix can't have been validated on the OV02C10 hardware it affects. That OV02C10 matrix is the shipped default for all Book3/Book4 users (via `webcam-fix-libcamera`), and it's intentionally near-identity because the sensor is uncalibrated upstream; users dial in their own via `tune-ccm.sh`. Changing the global default from an unrelated device risks regressing colour for every existing OV02C10 user. If you have a genuinely better OV02C10 matrix, it's very welcome — but as its own PR with OV02C10-hardware confirmation, or as a `tune-ccm.sh` preset rather than the shipped default. Everything else here is orthogonal to "Book5 NixOS + Bayer v0.7."

**2. The `camera-relay-monitor.c` change reintroduces `poll()` into the frame path**, which is the exact thing commit 9525152 deliberately removed because `poll()` between `read_full()`/`write()` caused periodic 30fps stutter. Your logic is careful (black-frame pump only fires on `poll()` timeout with no partial frame in flight, and during normal 30fps streaming frames arrive well inside the 100ms window so it never triggers) and the OBS-keepalive goal is legitimate — but because it re-touches a hot-loop regression we already paid for once, I want to run a 30fps no-stutter test on real hardware before merging. I'll do that on my end.

**Follow-up (not blocking this PR):** the v0.7 Bayer logic only reaches NixOS users, since the `.patch` files are consumed by the Nix overlay while the non-NixOS installer applies the fix through the embedded Python patcher in `webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh`, which this PR doesn't update. Not a regression (their path is unchanged), but I'll open an issue to port the "keep videoFormat, feed flipped order to ISP only" logic into the shell installer so Ubuntu/Fedora/Arch get it too.

Also heads-up: #58 (repo rename) just merged, so you'll hit a trivial README conflict — a quick rebase onto `main` will clear it. Thanks again!

Andycodeman (changes requested):
Thanks — both of the blocking notes are properly addressed, and I checked the code rather than the thread:

- **`ov02c10.yaml` CCM edits: dropped.** Confirmed *both* copies (`webcam-fix-book5/` and `webcam-fix-libcamera/`) are byte-identical to `main` now, so nothing snuck through in only one of them. 👍
- **Rebase: done**, README conflict from #58 is gone, GitHub says `MERGEABLE`.
- **`bayer-fix-v0.7.patch`**: I test-applied it against upstream libcamera **v0.6.0 and v0.7.0** — clean on both, so the Nix overlay is safe whichever version nixpkgs pins. The `videoFormat`/`ispFormat` split is exactly the right shape for the IPU7 link-validation EPIPE.
- **`camera-relay-monitor.c`: the stutter concern is dead — you were right.** I said I'd test this on hardware, then realised I didn't need to: the frame loop reads a *pipe* and writes an *fd*, so it's testable anywhere. I added `camera-relay/tests/relay-loop-bench.c`, which drives both the old blocking loop and your poll loop with a synthetic 30 fps 1080p YUY2 producer:

  ```
  30fps streaming — the poll loop must not stutter (baseline: blocking read)
    blocking       frames=149  mean= 33.33ms  p99= 35.31ms  max= 35.86ms  late(>40ms)=0
    poll loop      frames=149  mean= 33.33ms  p99= 36.78ms  max= 36.86ms  late(>40ms)=0

  2500ms pipeline startup — the device must not go silent past OBS's timeout
    blocking       black_frames=0    longest_silence=  2506ms  OBS(166ms) -> DISCONNECTS
    poll loop      black_frames=24   longest_silence=   102ms  OBS(166ms) -> stays connected
  ```

  Indistinguishable at 30 fps, and your pump does exactly what you said it does. For the record on why 9525152 doesn't repeat here: *that* `poll()` sat between `read_full()` and `write()`, adding a wakeup to every frame's delivery. Yours sits before each read, where the thread was going to block anyway — so it costs one extra syscall per 64 KB pipe chunk and adds nothing to delivery latency. Good instinct putting it there. The bench fails non-zero on either regression, so this is now a standing test rather than something a maintainer has to remember to eyeball.
- **NixOS module**: dropping `RELAY_COLOR_FILTER=videoflip` when `videoFlip=true` is right, and I agree with your reasoning — both v0.6 and v0.7 compose `Transform::Rot180`, so libcamera already flips the sensor and the old relay `videoflip` was a second (and only VFlip-partial) flip on top. `relayColorFilter` as an opt-in string is a nice generalisation.

**One thing I need reverted before merge: `webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh`.**

I know I flagged the shell-installer port as a follow-up, and I appreciate you taking a run at it — but it needs to land as its own PR, because it changes behaviour for users neither of us can test from here. NixOS consumes the `.patch` through the Nix overlay and never runs this script; this script is the shipped path for every **Ubuntu / Fedora / Arch** Book5 user.

I ran both patchers against real libcamera sources, and they emit **different Bayer orders**:

| | ISP Bayer order | for `Rot180` |
|---|---|---|
| `main` today | `native ^ (combinedTransform & 1)` — HFlip bit **only** | `native ^ 1` |
| this PR | `sensor_->bayerOrder(...)` → `BayerFormat::transform()`, XORs bit0 on HFlip **and** bit1 on VFlip | `native ^ 3` |

That `^1` isn't arbitrary — it's empirically derived (the comment in `main` records that OV02E10 only shifts its Bayer pattern on HFlip, *not* VFlip), and it's the version confirmed working on **940XHA/Fedora** and **960XHA/Ubuntu**. Switching those users to `^3` untested is a straight purple-tint regression risk for the existing base. It may well turn out that `^3` is the correct value and the old comment is wrong — but that's a question to settle *with OV02E10 hardware running the shell installer*, which is precisely what the follow-up issue is for.

The same hunk also drops two things from the primary patch path (they survive only in the legacy v0.5 branch):

1. the `LIBCAMERA_BAYER_ORDER=0..3` manual override — the escape hatch when auto-detection picks the wrong order;
2. the `[BAYER-FIX] transform=… origOrder=… newOrder=…` log line — which is the line I ask users to grep for to prove the patched libcamera is actually loaded (see #49). The `[BAYER-FIX] dispatch:` log survives, but not the order one.

So:

```bash
git checkout origin/main -- webcam-fix-book5/libcamera-bayer-fix/build-patched-libcamera.sh
git commit -m "revert: keep shell installer's bayer patcher out of this PR"
git push
```

…and I'll merge as soon as that lands — nothing else is outstanding. I'll open the shell-installer port as its own issue and tag you on it, with the `^1`-vs-`^3` question written up, since your v0.7 work is what makes it answerable. If you want to take that one too it's yours; it just needs an OV02E10 tester on a non-NixOS distro to settle which XOR is actually right.

Really nice work on this — the v0.7 patch and the OBS keepalive are both solid. Thanks for sticking with it.

Andycodeman:
Correction to my own review — I need to walk one thing back.

I said dropping `RELAY_COLOR_FILTER = "videoflip method=vertical-flip"` from `videoFlip = true` was right, reasoning that the forced `Rot180` reaches the sensor's flip controls (`setupFormats()` → `sensor_->setFormat(format, transform)`), so `libcamerasrc` would already emit an upright frame and the relay flip would be a second, redundant flip.

**But your own testing says otherwise, in two places:**

- Your PR description: *"Previously, only `LIBCAMERA_FORCE_OV02E10_ROTATION` was being passed (which fixed the colors), but the geometric flip was omitted, leaving the V4L2 output physically upside down."*
- #59: the V4L2 source is right-side up **with** `RELAY_COLOR_FILTER=videoflip method=vertical-flip`.

If that's still true, then after this PR a user who sets `videoFlip = true` and nothing else gets **correct colours but an upside-down image on the V4L2/relay path** — because `relayColorFilter` now defaults to `""`. PipeWire-native apps would still look fine (they honour the orientation metadata libcamera reports; the raw `v4l2sink` relay path doesn't), which would neatly explain why the two paths disagree — and it's consistent with `CameraSensorLegacy::setFormat()` only applying the flip controls `if (supportFlips_)`.

I can't settle this from the source, and I don't have the hardware. You do, and it's a 30-second check:

> With your final PR state — `videoFlip = true`, `relayColorFilter` **unset** — is the image in OBS (or any V4L2 app) right-side up, or upside down?

- **Right-side up** → I was right, the flip was redundant, nothing to change. 👍
- **Upside down** → this is an upgrade regression for existing `videoFlip = true` users, and we should either keep the relay flip on `videoFlip` by default, or at minimum say so loudly in the `videoFlip` option description so people know to set `relayColorFilter` too.

Everything else in my review stands: the one file to revert is still `build-patched-libcamera.sh`. Sorry for the extra round trip — better a question now than an upside-down camera for every NixOS user who upgrades.

Andycodeman:
Small correction: I referred to the shell-installer port as #69, but Andy had already opened **#62** for it back on 2026-07-03. I've closed #69 as a duplicate and moved the `^1`-vs-`^3` analysis over to **#62** — that's the one to follow.

Andycodeman (approved):
This is ready. Thanks for the thorough follow-up — you addressed the blocker and improved on the thing I was most worried about.

**Verified before approving:**

1. **`build-patched-libcamera.sh` is reverted out** — it's now byte-identical to `main` (confirmed against current `main`, which has moved on since the review). The blocker is cleared.

2. **The `^1`-vs-`^3` Bayer concern is resolved — better than I asked.** The OV02E10 path now uses the HFlip-only XOR (`cfgBayer.order ^ (combinedTransform() & 1)`), matching the `^1` behaviour confirmed on 940XHA/960XHA, and restricts the generic `bayerOrder()` (`^3`) to other sensors. So the NixOS `.patch` and the shell installer now agree on OV02E10. I re-tested the modified patch: it still applies cleanly to libcamera 0.7.0 and 0.7.2.

3. **On the orientation decoupling — I need to walk back my own correction.** I flagged that dropping `RELAY_COLOR_FILTER` from `videoFlip` might leave V4L2 output upside-down. I checked the OV02E10 kernel driver, and it registers both `V4L2_CID_HFLIP` and `V4L2_CID_VFLIP`, so libcamera sets `supportFlips_ = true` and your forced `Rot180` flows through `setFormat()` into the sensor's actual flip controls — the sensor rotates in hardware and `libcamerasrc` emits an upright frame. So the relay flip really was redundant, my *original* review call was right, and my correction was the mistake. `relayColorFilter` as a separate opt-in is the correct design. Sorry for the round trip.

4. **The `nixpkgsUnpatched` cascade escape hatch is a genuinely nice addition** — pinning pipewire back to break the overlay's rebuild cascade through every libpipewire consumer is exactly right, and the comment explaining why `prev.pipewire` is already tainted is appreciated.

**One note for the release, not a blocker:** gating the module behind a new `enable` option (`config = lib.mkIf cfg.enable`) is correct NixOS convention, but it's a breaking change for anyone already importing the module — they'll need to add `enable = true`. I'll call that out in the release notes.

Merging. Thanks again — the v0.7 Bayer work and the OBS keepalive are both solid contributions.

PR #75, opened 2026-07-29
Fix alc298 amp init service for Book3 Pro 14

vaibhavagg101:
Previously, the service was failing on boot. This happened because systemd would attempt to run the script before the kernel had fully prepared the sound card. Systemd will now actively wait for the specific audio device node (/dev/snd/hwC0D0) to appear before it tries to execute the script.

### DIFF
--- a/speaker-fix-940xfg/99-alc298-amp-init.rules
+++ b/speaker-fix-940xfg/99-alc298-amp-init.rules
@@ -1,0 +1,11 @@
+# Samsung Galaxy Book3 Pro 14" (940XFG) — start the ALC298 amp init as soon as
+# the sound card is fully registered.
+#
+# Why controlC? and not hwC?D?: /dev/snd/hwC*D* is never tagged for systemd
+# (99-systemd.rules tags controlC* only), which is what made the .device
+# dependency in this PR unsatisfiable. controlC* is created by
+# snd_card_register(), i.e. after every codec's hwdep node has been built, so
+# it is a guaranteed-late trigger with no card-index assumption baked in.
+# alc298-amp-init.sh does the real vendor_id/subsystem_id matching and exits 0
+# when the codec isn't ours, so firing on a second sound card is harmless.
+ACTION=="add", SUBSYSTEM=="sound", KERNEL=="controlC?", TAG+="systemd", ENV{SYSTEMD_WANTS}+="alc298-amp-init.service"
--- a/speaker-fix-940xfg/README.md
+++ b/speaker-fix-940xfg/README.md
@@ -43,7 +43,7 @@ This is **not** a kernel module. It does not patch `snd-hda-codec-realtek`.
 It is a userspace fix:
 
 - One bash script that talks to the codec via `hda-verb` (from `alsa-tools`)
-- One systemd unit that runs the script at boot
+- One systemd unit, activated by udev when the sound card registers (not enabled at boot)
 - One `system-sleep` hook that re-runs it after resume from suspend
 
 That keeps the fix entirely outside the kernel module hierarchy: it survives
@@ -74,7 +74,7 @@ The installer will:
 1. Verify the DMI product is `940XFG` and a codec with SSID `0x144dc882` is present.
 2. Install `alsa-tools` if `hda-verb` is missing.
 3. Copy `alc298-amp-init.sh` to `/usr/local/sbin/`.
-4. Install and enable `alc298-amp-init.service` (runs at every boot).
+4. Install `alc298-amp-init.service` and the udev rule that starts it as soon as the sound card registers (`controlC*` node appears) — not enabled at boot.
 5. Install `/lib/systemd/system-sleep/alc298-amp-init` (re-runs after suspend/resume).
 6. Fire the script once so speakers work immediately, no reboot required.
 
--- a/speaker-fix-940xfg/alc298-amp-init.service
+++ b/speaker-fix-940xfg/alc298-amp-init.service
@@ -1,14 +1,16 @@
 [Unit]
 Description=Initialize ALC298 internal speaker amps (Samsung Galaxy Book3 Pro 14" / 940XFG)
 Documentation=https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/issues/44
-Requires=dev-snd-hwC0D0.device
-After=systemd-modules-load.service sound.target dev-snd-hwC0D0.device
-Wants=sound.target 
+After=systemd-modules-load.service
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/local/sbin/alc298-amp-init.sh
 
-[Install]
-WantedBy=multi-user.target
+# Deliberately no [Install] section — udev starts this unit when the sound card
+# registers (see /etc/udev/rules.d/99-alc298-amp-init.rules). Adding
+# WantedBy=multi-user.target back would reintroduce the original bug: the boot
+# transaction can start the unit before the codec exists, the script no-ops,
+# and RemainAfterExit=yes then leaves it active so udev's later trigger does
+# nothing at all.
--- a/speaker-fix-940xfg/alc298-amp-init.service
+++ b/speaker-fix-940xfg/alc298-amp-init.service
@@ -1,9 +1,9 @@
 [Unit]
 Description=Initialize ALC298 internal speaker amps (Samsung Galaxy Book3 Pro 14" / 940XFG)
 Documentation=https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/issues/44
-After=systemd-modules-load.service sound.target
-Wants=sound.target
-ConditionPathExists=/dev/snd/hwC0D0
+Requires=dev-snd-hwC0D0.device
+After=systemd-modules-load.service sound.target dev-snd-hwC0D0.device
+Wants=sound.target 
 
 [Service]
 Type=oneshot
--- a/speaker-fix-940xfg/install.sh
+++ b/speaker-fix-940xfg/install.sh
@@ -131,8 +131,11 @@ echo "Installing system-sleep resume hook..."
 mkdir -p /lib/systemd/system-sleep
 install -m 755 "${SCRIPT_DIR}/alc298-amp-init-sleep.sh"  /lib/systemd/system-sleep/alc298-amp-init
 
+echo "Installing udev rule..."
+install -m 644 "${SCRIPT_DIR}/99-alc298-amp-init.rules" /etc/udev/rules.d/99-alc298-amp-init.rules
+
 systemctl daemon-reload
-systemctl enable alc298-amp-init.service
+udevadm control --reload-rules
 
 # ---------------------------------------------------------------------------
 # Fire once now so the user has working speakers without rebooting
--- a/speaker-fix-940xfg/uninstall.sh
+++ b/speaker-fix-940xfg/uninstall.sh
@@ -17,8 +17,10 @@ echo "Removing installed files..."
 rm -f /usr/local/sbin/alc298-amp-init.sh
 rm -f /etc/systemd/system/alc298-amp-init.service
 rm -f /lib/systemd/system-sleep/alc298-amp-init
+rm -f /etc/udev/rules.d/99-alc298-amp-init.rules
 
 systemctl daemon-reload
+udevadm control --reload-rules
 
 echo ""
 echo "=== Uninstall complete ==="


Andycodeman (changes requested):
Thanks for this, and thanks for digging into the ordering rather than just
working around it — the problem you've identified is real. But I can't take
this patch as written, because `dev-snd-hwC0D0.device` never actually exists as
far as systemd is concerned, so it makes the symptom worse rather than better.

systemd only creates an *active* `.device` unit for udev devices tagged
`systemd`, and the only sound rule that adds that tag is in
`99-systemd.rules`:

```
SUBSYSTEM=="sound", KERNEL=="controlC*", TAG+="systemd", ENV{SYSTEMD_WANTS}+="sound.target"
```

That matches `controlC*` only — `hwC*D*` never gets tagged. On a live ALC298
box here, with the node sitting right there in `/dev/snd`:

```
$ udevadm info /dev/snd/hwC0D0 | grep TAGS
E: TAGS=:uaccess:
$ systemctl show dev-snd-hwC0D0.device -p ActiveState
ActiveState=inactive
```

I built a throwaway unit with your exact `[Unit]` block to be sure. The job
sits in `waiting` while the device job spins, and after
`DefaultDeviceTimeoutSec` it gives up:

```
$ time systemctl start pr75-boottime.service
A dependency job for pr75-boottime.service failed.
exit=1 after 91s
$ journalctl -t pr75-boottime
-- No entries --
```

So the amp init would never run at boot at all, and you'd eat ~90s of hung job
on the way. `systemd-analyze verify` is happy with the file, which is why this
doesn't show up until you actually boot it.

The race you're describing is real though. `After=sound.target` isn't the
guarantee it looks like — `sound.target` is passive and can already have been
reached before the card registers, in which case the script runs, finds no
codec with SSID `0x144dc882`, exits 0, and `RemainAfterExit=yes` leaves the
unit sitting `active (exited)` having done nothing. Silent speakers, no failed
unit.

The one-line version of what you want is to hang the dependency off the
*control* device, which is tagged, and which ALSA creates only after every
codec's hwdep node is built:

```ini
Wants=sound.target dev-snd-controlC0.device
After=systemd-modules-load.service sound.target dev-snd-controlC0.device
ConditionPathExists=/dev/snd/hwC0D0
```

`Wants=` rather than `Requires=` so a renumbered card degrades instead of
hard-failing, and please keep the `ConditionPathExists=` line — it's what makes
the unit skip cleanly instead of going red.

Tested that shape here: starts in 1s, `ExecStart` runs. Could you try it on
your 940XFG and confirm the speakers come up from a cold boot? If it works I'll
merge it.

If you'd rather do it properly, the more robust option is a udev rule that
matches the codec by ID and starts the service directly — that also survives
card renumbering, which the hardcoded `hwC0D0` doesn't. Happy to write that up
if you want to go that route; the one gotcha is you'd need to drop
`WantedBy=multi-user.target` so the early no-op can't beat udev to it.

Last tiny thing: there's a trailing space on `Wants=sound.target ` — harmless,
but worth dropping while you're in there.

Also, could you attach `journalctl -b -u alc298-amp-init.service` from a bad
boot? I'd like to know whether you were seeing the condition skip it or the
run-and-no-op case, so the fix is aimed at the right one.

vaibhavagg101:
I’m afraid I've lost the original failure logs. I only had the logs from my own patch failing, which confirms your 90-second timeout diagnosis. It appears my code worked only once and hastening to commit it was a mistake on my end.

I have applied your suggested changes to the service file (and dropped the trailing space). I did multiple cold boots after the changes and the speakers came up immediately on every boot. But I’m still not sure if it’s a universal fix for all previous and future users who use this fix.

Assuming we have the logic sorted out, I would absolutely like to take you up on the offer of a udev rule for the proper fix. I'd prefer not to push a workaround that I don't fully grasp. Let me know what you need from me to get that set up.

Andycodeman:
That's genuinely useful, thanks — and don't worry about the lost logs. You confirming the 90s timeout from your own boot is the important half, and "it worked once" is exactly the shape of a race, so I think we've got it pinned.

**One housekeeping thing first:** the branch here is still on `62cdaffb` with the original `Requires=dev-snd-hwC0D0.device`, so your local fix hasn't been pushed. Nothing to fix on your end yet — just flagging it so neither of us merges this by accident.

And your multiple cold boots on the 940XFG are the piece I couldn't get from here (I've only got an ALC298 machine, not a 940XFG), so that's the hardware confirmation this needed. Thank you for actually doing the boots.

Your instinct about "is it universal?" is right, and it's the reason to go to the udev version. `dev-snd-controlC0.device` hardcodes card index 0, and so does `ConditionPathExists=/dev/snd/hwC0D0`. Plug in a USB DAC or a dock that enumerates first and your ALC298 becomes card1 — at which point the unit waits on the wrong card. `alc298-amp-init.sh` already scans every `hwC*D*` by vendor + SSID, so the *script* survives renumbering; it's only the unit wiring that doesn't. Let's fix the wiring to match.

---

## The udev version

### 1. New file: `speaker-fix-940xfg/99-alc298-amp-init.rules`

```
# Samsung Galaxy Book3 Pro 14" (940XFG) — start the ALC298 amp init as soon as
# the sound card is fully registered.
#
# Why controlC? and not hwC?D?: /dev/snd/hwC*D* is never tagged for systemd
# (99-systemd.rules tags controlC* only), which is what made the .device
# dependency in this PR unsatisfiable. controlC* is created by
# snd_card_register(), i.e. after every codec's hwdep node has been built, so
# it is a guaranteed-late trigger with no card-index assumption baked in.
# alc298-amp-init.sh does the real vendor_id/subsystem_id matching and exits 0
# when the codec isn't ours, so firing on a second sound card is harmless.
ACTION=="add", SUBSYSTEM=="sound", KERNEL=="controlC?", TAG+="systemd", ENV{SYSTEMD_WANTS}+="alc298-amp-init.service"
```

I did consider matching the codec directly on `hwC?D?` with `ATTR{vendor_id}=="0x10ec0298", ATTR{subsystem_id}=="0x144dc882"`, and it does work — I tested it with `udevadm test` and it matches the ALC298 while correctly ignoring the Intel HDMI codec on the same card. I backed off it because it depends on those sysfs attributes already existing at the moment the `add` uevent fires, and I can't verify that ordering from userspace without a reboot I can't stage. If it were ever false we'd be shipping you a rule that silently never fires — the same class of bug we're fixing. `controlC?` has a guarantee I can actually point at, so let's take the boring one.

Verified the composition on my machine — our rule and systemd's own `controlC*` rule both use `+=`, so they accumulate rather than clobber:

```
$ udevadm test /sys/class/sound/controlC0
SYSTEMD_WANTS=alc298-amp-init.service sound.target
TAGS=:systemd:uaccess:
```

### 2. `speaker-fix-940xfg/alc298-amp-init.service` — replace the whole file

```ini
[Unit]
Description=Initialize ALC298 internal speaker amps (Samsung Galaxy Book3 Pro 14" / 940XFG)
Documentation=https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/issues/44
After=systemd-modules-load.service

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/local/sbin/alc298-amp-init.sh

# Deliberately no [Install] section — udev starts this unit when the sound card
# registers (see /etc/udev/rules.d/99-alc298-amp-init.rules). Adding
# WantedBy=multi-user.target back would reintroduce the original bug: the boot
# transaction can start the unit before the codec exists, the script no-ops,
# and RemainAfterExit=yes then leaves it active so udev's later trigger does
# nothing at all.
```

That last paragraph is the one real trap in this approach, so here it is measured rather than asserted — a oneshot with `RemainAfterExit=yes`, started twice:

```
$ systemctl start alc298-trap-probe.service   # exit=0
$ systemctl start alc298-trap-probe.service   # exit=0
$ journalctl -t alc298-trap | grep -c RAN
1                                             # <-- ran once, not twice
ActiveState=active SubState=exited
```

Both starts "succeed"; the second does nothing. So `[Install]` has to go, not just be reordered.

### 3. `speaker-fix-940xfg/install.sh`

Add the rule next to the other installs:

```bash
echo "Installing udev rule..."
install -m 644 "${SCRIPT_DIR}/99-alc298-amp-init.rules" /etc/udev/rules.d/99-alc298-amp-init.rules
```

and swap the enable for a rules reload:

```diff
 systemctl daemon-reload
-systemctl enable alc298-amp-init.service
+udevadm control --reload-rules
```

(`systemctl enable` on a unit with no `[Install]` exits 0 but prints a "not meant to be enabled" paragraph at the user, so it's worth removing rather than leaving in — it won't trip the script's `set -e`, it just looks broken.)

The existing "fire once now" block at the end still does the right thing, so no reboot needed after install.

### 4. `speaker-fix-940xfg/uninstall.sh`

```diff
 rm -f /usr/local/sbin/alc298-amp-init.sh
 rm -f /etc/systemd/system/alc298-amp-init.service
 rm -f /lib/systemd/system-sleep/alc298-amp-init
+rm -f /etc/udev/rules.d/99-alc298-amp-init.rules
 
 systemctl daemon-reload
+udevadm control --reload-rules
```

The existing `systemctl disable ... || true` is already guarded, so it's fine to leave as-is.

### 5. `speaker-fix-940xfg/README.md`

Two lines go stale: "One systemd unit that runs the script at boot" (line 46) and "Install and enable `alc298-amp-init.service` (runs at every boot)" (line 77). Worth saying it's udev-activated on card registration rather than enabled at boot, so the next person reading it doesn't go looking for a `WantedBy`.

---

## What I'd like from you

Push all of that to this branch, then do the same thing you did last time — a couple of cold boots, and confirm the speakers come up. The check that matters:

```bash
systemctl status alc298-amp-init.service
journalctl -t alc298-amp-init -b
```

You want `active (exited)` and the `Initialized ALC298 internal speaker amps on /dev/snd/hwC0D0` line, on every boot, without you touching anything. If you've got a USB audio device or a dock handy, a boot with it plugged in is a nice bonus test — that's the case the old `hwC0D0` wiring would have got wrong.

Once that's confirmed on your hardware I'll merge it. And to be clear about what's tested where: everything above about systemd and udev mechanics I've verified here, but the amp COEF sequence and the actual speakers are yours — I have no 940XFG to check against.

No rush, and thanks for sticking with this rather than letting the first version land.

vaibhavagg101:
Tested on my machine. Before switching over, I disabled the old boot-enabled unit while its `[Install]` section was still on disk, so the switch to udev-only activation wouldn't leave a stale `multi-user.target.wants` symlink racing the new rule:
    `sudo systemctl disable --now alc298-amp-init.service`
Then re-ran `install.sh`

Ran 4 full cold boots. Speakers came up every time. The suggested commands confirm the unit is now static, and shows exactly one clean run per boot with an identical COEF sequence each time:

```
vaibhav@vaibhav-940XFG:~$ systemctl status alc298-amp-init.service
journalctl -t alc298-amp-init -b
● alc298-amp-init.service - Initialize ALC298 internal speaker amps (Samsung Galaxy Book3 Pro 14" / 940XFG)
     Loaded: loaded (/etc/systemd/system/alc298-amp-init.service; static)
     Active: active (exited) since Wed 2026-08-05 02:48:48 IST; 45s ago
 Invocation: d014fdfd4cbc45dea6b5eb928accb725
       Docs: https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/issues/44
    Process: 2419 ExecStart=/usr/local/sbin/alc298-amp-init.sh (code=exited, status=0/SUCCESS)
   Main PID: 2419 (code=exited, status=0/SUCCESS)
   Mem peak: 6.3M
        CPU: 715ms

Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3869]: nid = 0x20, verb = 0x4b0, param = 0x11
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3870]: nid = 0x20, verb = 0x500, param = 0x23
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3875]: nid = 0x20, verb = 0x423, param = 0x9e
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3876]: nid = 0x20, verb = 0x500, param = 0x25
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3880]: nid = 0x20, verb = 0x400, param = 0x4
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3882]: nid = 0x20, verb = 0x500, param = 0x26
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3887]: nid = 0x20, verb = 0x4b0, param = 0x11
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3889]: nid = 0x17, verb = 0x707, param = 0x40
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init.sh[3893]: nid = 0x17, verb = 0x300, param = 0xb000
Aug 05 02:48:48 vaibhav-940XFG systemd[1]: Finished alc298-amp-init.service - Initialize ALC298 internal speaker amps (Samsung Galaxy Book3 Pro 14" / 940XFG).
Aug 05 02:48:48 vaibhav-940XFG alc298-amp-init[3895]: Initialized ALC298 internal speaker amps on /dev/snd/hwC0D0
vaibhav@vaibhav-940XFG:~$ 
```

Same pattern on all 4 boots, only the PID/timestamp differ.

Andycodeman (approved):
Merging. Thanks for sticking with this — three rounds and a rewrite is a lot to ask of a first contribution, and you were right at the start that there was a real bug here.

**Your evidence is exactly the right evidence.** `Loaded: … static` is the line that matters: it proves the unit has no `[Install]` section, which is what makes udev the *only* trigger. That's the property the whole design depends on, and it's the one that would silently rot if someone later re-added `WantedBy=multi-user.target`. Four cold boots, one clean run each, identical COEF sequence, and the `Initialized ALC298 internal speaker amps on /dev/snd/hwC0D0` line at the end.

Disabling the old unit *before* switching, so a stale `multi-user.target.wants` symlink couldn't race the new rule, was a good catch and not something I'd flagged. That would have been a confusing bug to chase — two activations, one of them early and no-opping.

Checked here: `systemd-analyze verify` clean, `bash -n` clean on all four scripts, `uninstall.sh`'s `systemctl disable` already carries `2>/dev/null || true` so it won't print the "not meant to be enabled" paragraph now that there's no `[Install]`.

One thing I verified rather than assumed, because it would have been a silent killer: the rules file has **no trailing newline**, and it contains exactly one rule — so if udev skipped an unterminated final line, nothing would load at all and your four boots would have been luck. It doesn't. A byte-identical copy parses fine and composes with systemd's own `controlC*` rule rather than clobbering it:

```
last byte: 22 (not 0a)
SYSTEMD_WANTS=alc298-amp-init.service sound.target
```

## One follow-up, not blocking

The `RemainAfterExit` trap your unit comment warns about isn't quite gone — it's moved rather than removed. The comment describes the boot transaction beating udev to it; the same thing can happen with a **second sound card**.

If another card's `controlC*` registers before the HDA card's, the service fires early, `alc298-amp-init.sh` scans and finds no codec with SSID `0x144dc882`, exits 0 — and `RemainAfterExit=yes` leaves the unit `active (exited)`. The later trigger for the real card then does nothing, because systemd sees it as already started. Silent speakers, no failed unit. Exactly the shape you've documented, just from a different direction.

I'm not holding the merge for it: on a 940XFG the PCI HDA card enumerates before USB audio, the installer is DMI+SSID gated, and your four cold boots are real. But it's cheap to close if you want to:

```diff
 Type=oneshot
-RemainAfterExit=yes
```

`alc298-amp-init.sh` already documents itself as idempotent ("safe to repeat"), so letting the unit go inactive means every `controlC*` add re-runs it and no ordering can strand it. The only cost is that `systemctl status` reads `inactive (dead)` between boots instead of `active (exited)`, which is normal for a fire-and-forget oneshot.

Happy to take that as a follow-up PR, or leave it — your call, and no rush either way.

Tiny thing: the `.rules` file is committed mode `755`. Harmless, since `install.sh` installs it `-m 644`.

Cutting a release with this in it. And for the record of how this went: you reported a real boot race, your first patch had the wrong mechanism, you tested my correction and found it worked, then asked for the proper fix rather than shipping the workaround you didn't fully follow. That last decision is the reason this ended up right.

Andycodeman:
Merged as `53e4172` and shipped in **v0.3.60**.

One thing I pushed on top (`4dc3825`), because it would have bitten every existing user: **your `systemctl disable --now` step wasn't optional.**

You mentioned doing it by hand so a stale `multi-user.target.wants` symlink wouldn't race the new rule. `install.sh` never did it for anyone else — and dropping `[Install]` from the unit doesn't remove that symlink. `[Install]` only tells `enable`/`disable` what to do; the symlink is what actually activates. Verified:

```
unit has [Install]?          0
systemctl show -p WantedBy:  multi-user.target      ← still activated
```

So on upgrade the unit would still have been pulled into the boot transaction, run before the codec existed, no-opped, and `RemainAfterExit=yes` would have left it active — after which the udev trigger does nothing. The exact bug we just removed, reintroduced for everyone who already had the fix installed.

`install.sh` now runs the disable explicitly. Confirmed it clears the symlink even with no `[Install]` present:

```
before disable -> WantedBy: multi-user.target
after  disable -> WantedBy: ''       symlink gone
```

Worth saying plainly: you found that one, I just noticed it needed to be in the installer rather than in your shell history. Mentioning the manual step in your test write-up is what made it visible — if you'd quietly done it and not said so, this would have shipped broken for every existing 940XFG user.

The `RemainAfterExit` follow-up from my review is noted in the release as a known gap, so it's recorded either way. No pressure on it.

Issue #70, opened 2026-07-13
Installer drift: webcam-fix-libcamera and webcam-fix-book5 diverge silently (Book5 is missing the NVIDIA black-camera fix)


Andycodeman:
`webcam-fix-libcamera/install.sh` (Book3/Book4) and `webcam-fix-book5/install.sh` (Book5) carry large amounts of near-duplicate logic. Fixes keep landing in one and not the other, and the only symptom users see is "the camera doesn't work."

This has now bitten twice:

- **#49 → #51.** The Fedora RPM Fusion / `akmod-v4l2loopback` fix landed in the Book5 installer in v0.3.42 and was never ported. The Book3/Book4 installer kept calling `dnf install -y v4l2loopback` — a package that doesn't exist in Fedora — with `2>/dev/null || true` swallowing the failure. Every Fedora Book3/Book4 user got a silently-broken camera. Fixed in v0.3.51.
- **#60.** A review had to explicitly check that a change to `ov02c10.yaml` landed in *both* copies, because a fix applied to one copy and not its sibling looks correct in review and is still broken.

## Audit: what's still out of sync

After the v0.3.51 fix, the RPM Fusion / akmods / OEM-v4l2-relayd / udev-capabilities / Secure-Boot blocks are in both. These are still **only in `webcam-fix-libcamera`**, and absent from `webcam-fix-book5`:

| Missing from Book5 installer | Origin | Does it apply to Book5? |
|---|---|---|
| `LIBCAMERA_SOFTISP_MODE=cpu` on NVIDIA | #15 | **Yes — likely a real bug.** libcamera's `debayer_egl.cpp` is incompatible with NVIDIA's proprietary EGL (`glFrameBufferTexture2D error 36054`) → **black camera**. This is GPU/libcamera-specific, not sensor-specific, and Book5 Pro ships in RTX dGPU configs. The Book5 installer has *no* NVIDIA handling at all. |
| `usermod -aG kvm` | #50 | **Probably.** Without `kvm` group membership `/dev/udmabuf` is unusable → GPU debayer can't import the buffer → black frames. Same SoftISP path on Book5. |
| `check_sensor_helper` | #50 | Partly. Book5 patches the *system* libcamera rather than building into `/usr/local`, so the `/usr/local`-specific parts don't map 1:1 — but "is the OV02E10 sensor helper registered?" is still a real question (a missing helper means generic AGC → dark image). |
[...]

Issue #66, opened 2026-07-09
IPU6 OV02C10: Low framerate (~13-14 fps) on Galaxy Book3 Ultra


rodrigost23:
## Problem

Camera captures frames but at reduced framerate (~13-14 fps instead of expected 30 fps).

## Symptoms

- Camera is detected and functional via Pipewire in Firefox, Chromium, and qcam
- Frame capture works but at ~13-14 fps
- Video appears laggy due to low framerate
- Color and exposure tuning works correctly (tune-ccm.sh presets apply)
- Image is a bit grainy
- Only framerate is affected

## Logs

Capturing some frames:

```log
$ cam -c1 -C30 --file=/tmp/frame.raw
[2:41:32.274029503] [222980]  INFO Camera camera_manager.cpp:340 libcamera v0.7.0+dirty (2026-07-09T10:56:24-03)
[2:41:32.283928264] [222981] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02c10 0-0036': Unable to get rectangle 2 on pad 0/0: Inappropriate ioctl for device
[2:41:32.283952003] [222981]  WARN CameraSensor camera_sensor_legacy.cpp:402 'ov02c10 0-0036': The PixelArraySize property has been defaulted to 1928x1092
[2:41:32.283957207] [222981] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02c10 0-0036': Unable to get rectangle 1 on pad 0/0: Inappropriate ioctl for device
[2:41:32.283961711] [222981]  WARN CameraSensor camera_sensor_legacy.cpp:413 'ov02c10 0-0036': The PixelArrayActiveAreas property has been defaulted to (0, 0)/1928x1092
[2:41:32.283967546] [222981] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02c10 0-0036': Unable to get rectangle 0 on pad 0/0: Inappropriate ioctl for device
[2:41:32.283971159] [222981]  WARN CameraSensor camera_sensor_legacy.cpp:421 'ov02c10 0-0036': Failed to retrieve the sensor crop rectangle
[2:41:32.283974648] [222981]  WARN CameraSensor camera_sensor_legacy.cpp:427 'ov02c10 0-0036': The sensor kernel driver needs to be fixed
[...]

PR #76, opened 2026-08-03
camera-relay: fix black camera on hybrid-GPU laptops (EGL debayer lands on NVIDIA)

4nrry:
## The bug

On a hybrid-GPU laptop — every Galaxy Book4 Ultra, and the RTX variants of the Book5 Pro — the webcam privacy LED lights up, `camera-relay status` reports `STREAMING`, and apps get a black picture ("Your camera may be blocked" in Google Meet).

libcamera's Software ISP debayers Bayer→RGB on the GPU through EGL. Which GPU that is comes from GLVND, which loads the vendor ICDs in `/usr/share/glvnd/egl_vendor.d` in filename order — and NVIDIA's ships as `10_nvidia.json`, ahead of Mesa's `50_mesa.json`. So the debayer runs on NVIDIA's proprietary EGL driver, which it is not compatible with:

```
ERROR eGL egl.cpp:134 glFrameBufferTexture2D error 36054
ERROR Debayer debayer_egl.cpp:639 debayerGPU failed
```

The sensor powers up — hence the LED — but every frame dies in the conversion and nothing reaches v4l2loopback.

**It only breaks under systemd.** A desktop session normally has EGL already resolved to Mesa, so running `cam` from a terminal works and the bug looks like it isn't there. That is what makes it so easy to misdiagnose.

## Evidence

Galaxy Book4 Ultra, Kubuntu 26.04, kernel 7.0.0-28, libcamera 0.7.2, Mesa 26.0.3, NVIDIA 595.84:

```
__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/50_mesa.json  cam -c 1 -C5
  -> GL_RENDERER: Mesa Intel(R) Arc(tm) Graphics (MTL), 30.08 fps, zero errors

__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/10_nvidia.json cam -c 1 -C5
  -> hangs, dies at the timeout, not one frame
```

## The fix

`camera-relay` pins `__EGL_VENDOR_LIBRARY_FILENAMES` to the iGPU's vendor ICD, in the generated service unit and in `setup_environment`.

Pinning is conditional. Only a genuinely mixed NVIDIA-proprietary + Mesa setup gets a pin — on a single-GPU box it is redundant, and on an NVIDIA-only box pointing at a Mesa ICD that cannot drive the hardware would break a working camera. The target is derived from the render nodes actually present, not hardcoded to `50_mesa.json`, since the number is only a load-order hint and distros renumber it.

| topology | result |
| --- | --- |
| i915 / xe / amdgpu only | no pin |
| nvidia only | no pin — falls through to the CPU-debayer guard |
| i915 + nvidia | pin |
| amdgpu + nvidia | pin |
| xe + nvidia-drm | pin |
| i915 + nouveau | no pin (both Mesa, no ordering problem) |

This supersedes the CPU-debayer fallback on hybrid machines, so the two are now **exclusive**. Setting both left `LIBCAMERA_SOFTISP_MODE=cpu` winning (it touches no EGL at all) and the pin dead in the unit on exactly the machines it was written for, at a large cost in framerate (#66). CPU debayer still covers the case with no iGPU to fall back to.

## `camera-relay doctor`

When the debayer fails, every other check stays green — libcamerasrc loads, refcount is non-zero, service is active, `status` says STREAMING. That green wall is what made this hard to place. New **GPU / EGL debayer** section:

```
── GPU / EGL debayer ──
  render node: /dev/dri/renderD128 (driver i915)
  render node: /dev/dri/renderD129 (driver nvidia)
  EGL vendor ICD: /usr/share/glvnd/egl_vendor.d/10_nvidia.json (libEGL_nvidia.so.0)
  EGL vendor ICD: /usr/share/glvnd/egl_vendor.d/50_mesa.json (libEGL_mesa.so.0)
  topology: HYBRID (NVIDIA dGPU + Mesa-driven iGPU)
  vendor pin (this shell):   none
  vendor pin (service unit): /usr/share/glvnd/egl_vendor.d/50_mesa.json
  pipeline EGL (from /run/user/1000/camera-relay.log):
    EGL_VENDOR: Mesa Project
    GL_RENDERER: Mesa Intel(R) Arc(tm) Graphics (MTL)
  debayer throughput: Debayer processed 30 frames in 141265us, 4708 us/frame
```

and on a broken machine:

```
  → HYBRID GPU WITH NO PIN. GLVND loads the NVIDIA ICD first, so the
    Software ISP's EGL debayer runs on NVIDIA's driver and fails.
  → GPU DEBAYER FAILED — this is why no picture reaches the loopback: …
```

`NO FRAMES` / `BLANK FRAMES` now name the GPU cause instead of leaving the reader to connect two sections thirty lines apart.

One fix worth calling out: doctor reads **both** relay logs. The pipeline lives in a different process per mode — always-on writes to `camera-relay-gst.log` via `start_pipeline()`, while on-demand (the default, and what `enable-persistent` sets up) runs `gst-launch` under `camera-relay-monitor` and everything lands in `camera-relay.log`. Reading only the first missed libcamera's EGL banner *and* the debayer failure in the mode nearly everyone runs.

## Installers

Both installers already force CPU debayer globally when NVIDIA is the active renderer (#15, #50, #70). That misses the hybrid case: the desktop renders on the iGPU, so `glxinfo` reports Intel and no fix is applied, yet the relay service still lands on `10_nvidia.json`. They now run the same detection and report it in the install log.

`__EGL_VENDOR_LIBRARY_FILENAMES` is deliberately **not** written to `/etc/environment.d` like `LIBCAMERA_SOFTISP_MODE` — it steers every GL client on the machine, so a global setting would take NVIDIA offload away from games and everything else. It belongs on the camera units only.

Both installers get the identical block, with a "keep in sync" comment, until #70's shared `lib/common.sh` exists.

## Known gap (documented, not fixed)

Apps reading the camera through **PipeWire's libcamera source** — Chromium with `#enable-webrtc-pipewire-camera`, which 6c205ba established is *required* on libcamera 0.7+ — run the debayer inside `pipewire.service`, which has no pin, so they can still hit this. The README gives the `pipewire.service.d` drop-in as a workaround. Not shipped: pinning EGL inside PipeWire for every user is a wider blast radius than this fix needs, and it deserves its own change.

## Testing

Verified on the Book4 Ultra above: unit regenerated with the pin, manual drop-in removed, `REAL PICTURE` with 128 distinct byte values, debayer confirmed on the Arc iGPU. Detection logic exercised across all six topologies in the table. Every commit is independently `bash -n` clean.

Not tested: a hybrid box with `prime-select nvidia`. That configuration previously got CPU debayer and now gets the pin — the behaviour change is intentional (it is what #66 asked for) but a second tester would be welcome.

Also not tested: an AMD iGPU + NVIDIA machine. The detection covers it, but no such Samsung model was available.

## Note on issue numbers

This bug has no filed issue; I did not invent a number for it. The references to #15/#50/#66/#70 are to the genuinely related work.

### DIFF
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -935,6 +935,104 @@ cmd_doctor() {
         fi
     fi
 
+    # The Software ISP debayers on the GPU via EGL, and on a hybrid laptop GLVND
+    # hands it the NVIDIA driver, which it cannot use. Everything else in this
+    # report stays green while that happens, which is exactly what made this so
+    # hard to place — so name the GPU explicitly.
+    echo
+    echo "── GPU / EGL debayer ──"
+    local gst_log="$rt/camera-relay-gst.log"
+    local gpu_debayer_failed=false hybrid_unpinned=false
+    local nodes_seen=0 saw_nvidia=false saw_mesa_drv=false
+    local rn_base rn_drv
+    while read -r rn_base rn_drv; do
+        [[ -n "$rn_base" ]] || continue
+        nodes_seen=$((nodes_seen + 1))
+        echo "  render node: /dev/dri/$rn_base (driver $rn_drv)"
+        case "$rn_drv" in
+            nvidia|nvidia-drm) saw_nvidia=true ;;
+            unknown) ;;
+            *) saw_mesa_drv=true ;;
+        esac
+    done < <(render_node_drivers)
+    (( nodes_seen == 0 )) && echo "  render node: NONE — no /dev/dri/renderD*, the GPU debayer cannot run at all"
+
+    local vd vf
+    for vd in /etc/glvnd/egl_vendor.d /usr/share/glvnd/egl_vendor.d; do
+        [[ -d "$vd" ]] || continue
+        for vf in "$vd"/*.json; do
+            [[ -f "$vf" ]] || continue
+            echo "  EGL vendor ICD: $vf ($(grep -oP '"library_path"\s*:\s*"\K[^"]+' "$vf" 2>/dev/null | head -1 || echo '?'))"
+        done
+    done
+
+    # GLVND loads these in the order printed above, so an NVIDIA ICD sorting
+    # first is not a curiosity — it is the bug.
+    local want_pin="" unit_pin="" env_pin="${__EGL_VENDOR_LIBRARY_FILENAMES:-}"
+    want_pin=$(detect_egl_vendor_pin) || want_pin=""
+    if systemctl --user cat "$SERVICE_NAME" &>/dev/null; then
+        unit_pin=$(systemctl --user cat "$SERVICE_NAME" 2>/dev/null \
+            | grep -oP '^Environment=__EGL_VENDOR_LIBRARY_FILENAMES=\K.*' | tail -1) || unit_pin=""
+    fi
+    if $saw_nvidia && $saw_mesa_drv; then
+        echo "  topology: HYBRID (NVIDIA dGPU + Mesa-driven iGPU)"
+    elif $saw_nvidia; then
+        echo "  topology: NVIDIA only"
+    elif (( nodes_seen > 0 )); then
+        echo "  topology: single Mesa-driven GPU — no vendor pin needed"
+    fi
+    echo "  vendor pin (this shell):   ${env_pin:-none}"
+    echo "  vendor pin (service unit): ${unit_pin:-none}"
+    if [[ -n "$want_pin" && -z "$unit_pin" ]]; then
+        hybrid_unpinned=true
+        echo "  → HYBRID GPU WITH NO PIN. GLVND loads the NVIDIA ICD first, so the"
+        echo "    Software ISP's EGL debayer runs on NVIDIA's driver and fails."
+        echo "    Fix: camera-relay disable-persistent && camera-relay enable-persistent"
+        echo "    (or drop in Environment=__EGL_VENDOR_LIBRARY_FILENAMES=$want_pin)"
+    fi
+
+    # What the pipeline actually got, rather than what it should have got.
+    # Both logs, because the pipeline lives in a different process per mode:
+    # always-on runs gst-launch from start_pipeline() into camera-relay-gst.log,
+    # while on-demand — the default — runs it under camera-relay-monitor and
+    # everything lands in camera-relay.log. Reading only the first one missed
+    # libcamera's EGL banner, and the debayer failure, in the common mode.
+    local pipe_logs=() lf_egl
+    for lf_egl in "$gst_log" "$rt/camera-relay.log"; do
+        [[ -s "$lf_egl" ]] && pipe_logs+=("$lf_egl")
+    done
+    if (( ${#pipe_logs[@]} == 0 )); then
+        echo "  pipeline EGL: no relay log yet — start the relay and retry"
+    else
+        local egl_lines bench_line
+        egl_lines=$(grep -haoE '(EGL_VENDOR|EGL_VERSION|GL_VERSION|GL_RENDERER):.*' \
+            "${pipe_logs[@]}" 2>/dev/null | tail -4) || egl_lines=""
+        if [[ -n "$egl_lines" ]]; then
+            echo "  pipeline EGL (from ${pipe_logs[*]}):"
+            printf '%s\n' "$egl_lines" | sed 's/^/    /'
+        else
+            echo "  pipeline EGL: not reported in ${pipe_logs[*]}"
+            echo "    (CPU debayer, or the pipeline never got that far)"
+        fi
+        # Proof that frames are really going through the debayer, not just that
+        # it initialised — libcamera prints this once it has processed a batch.
+        bench_line=$(grep -ha 'Debayer processed' "${pipe_logs[@]}" 2>/dev/null | tail -1) || bench_line=""
+        [[ -n "$bench_line" ]] && echo "  debayer throughput: ${bench_line#*INFO }"
+        if grep -haq 'debayerGPU failed\|glFrameBufferTexture2D error' "${pipe_logs[@]}" 2>/dev/null; then
+            gpu_debayer_failed=true
+            echo "  → GPU DEBAYER FAILED — this is why no picture reaches the loopback:"
+            { grep -ha 'debayerGPU failed\|glFrameBufferTexture2D error' "${pipe_logs[@]}" 2>/dev/null \
+                | tail -3 | sed 's/^/    /'; } || true
+            echo "    The sensor still powers on (privacy LED lit) and the relay still"
+            echo "    reports STREAMING, but every frame is dropped in Bayer→RGB."
+            if [[ -n "$want_pin" ]]; then
+                echo "    Fix: pin EGL to the iGPU — $want_pin"
+            else
+                echo "    Fix: LIBCAMERA_SOFTISP_MODE=cpu in the service unit (slower, always works)"
+            fi
+        fi
+    fi
+
     echo
     echo "── v4l2loopback ──"
     # The module's reference count is the cheapest proof that something really
@@ -997,9 +1095,22 @@ cmd_doctor() {
             --stream-to=- 2>/dev/null | tail -c 262144 > "$frame" || true
         local size=0
         [[ -f "$frame" ]] && size=$(stat -c %s "$frame" 2>/dev/null || echo 0)
+        # A black picture with the relay "up" is the signature of the hybrid-GPU
+        # debayer failure, so say so here instead of leaving it to the reader to
+        # connect this section with the EGL one.
+        gpu_note() {
+            if $gpu_debayer_failed; then
+                echo "  → CAUSE: the GPU debayer failed (see 'GPU / EGL debayer' above)."
+                echo "    Nothing is wrong with the sensor or the loopback."
+            elif $hybrid_unpinned; then
+                echo "  → LIKELY CAUSE: hybrid GPU with no EGL vendor pin (see above)."
+                echo "    Re-run 'camera-relay enable-persistent' and check again."
+            fi
+        }
         if (( size == 0 )); then
             echo "  NO FRAMES — the device exists but nothing is writing to it."
             echo "  → the relay is not feeding the loopback. Check the logs below."
+            gpu_note
         else
             local distinct
             distinct=$(od -An -tu1 "$frame" 2>/dev/null | tr -s ' ' '\n' \
@@ -1008,12 +1119,14 @@ cmd_doctor() {
                 echo "  BLANK FRAMES — $size bytes captured, only $distinct distinct byte values."
                 echo "  → apps can open the device but the picture is uniform (black)."
                 echo "    The relay is holding the device open; the camera pipeline is not producing."
+                gpu_note
             else
                 echo "  REAL PICTURE — $size bytes captured, $distinct distinct byte values."
                 echo "  → the relay side is working end to end. If a browser still shows no"
                 echo "    camera, the problem is browser-side (see below), not the relay."
             fi
         fi
+        unset -f gpu_note
         rm -f "$frame"
     fi
 
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -1276,9 +1276,17 @@ cmd_enable_persistent() {
     # Deliberately exclusive. Setting both would leave CPU mode winning (it uses
     # no EGL at all) and the pin dead in the unit on exactly the hybrid machines
     # it was written for.
+    #
+    # Not choosing cpu is not enough to get gpu, though: the installers write
+    # LIBCAMERA_SOFTISP_MODE=cpu to /etc/environment.d for the PipeWire-libcamera
+    # path, systemd's 30-systemd-environment-d-generator feeds that into the user
+    # manager, and this service inherits it — which would make the pin inert
+    # exactly here. A unit-level Environment= outranks the inherited value, so
+    # the hybrid branch has to state gpu rather than merely stay silent.
     local egl_vendor=""
     egl_vendor=$(detect_egl_vendor_pin) || egl_vendor=""
     if [[ -n "$egl_vendor" ]]; then
+        softisp_mode="gpu"
         info "Hybrid GPU detected — service will pin EGL to $egl_vendor (GPU debayer on the iGPU)"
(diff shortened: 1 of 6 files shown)


Andycodeman (changes requested):
Thank you for this — it's an unusually good contribution, and I want to say so before I start picking at it. You found a bug nobody had filed, chased it to the actual mechanism instead of stopping at a workaround, tested it in both directions, and then wrote up the parts you *hadn't* tested. That last bit is rarer than it should be and it made this reviewable in an evening rather than a weekend.

The diagnosis is specific and falsifiable, the A/B with `__EGL_VENDOR_LIBRARY_FILENAMES` is exactly the evidence I'd have asked for, and the detection is written the way I'd want it — keyed on render nodes rather than on which vendor JSONs happen to be installed, and deriving the target ICD instead of hardcoding `50_mesa.json`.

I checked the detection rather than taking the table on trust. I pulled the three functions off the branch, mocked the render-node enumeration, and ran the real `find_mesa_egl_vendor` against this machine's ICDs:

```
i915 only            nopin  PASS      i915 + nvidia        pin    PASS
xe only              nopin  PASS      amdgpu + nvidia      pin    PASS
amdgpu only          nopin  PASS      xe + nvidia-drm      pin    PASS
nvidia only          nopin  PASS      i915 + nouveau       nopin  PASS
```

Plus four you didn't list — no render nodes, `nvidia + unknown`, two NVIDIA nodes, and `i915 + nvidia + unknown` — all behaving sensibly. 12/12. Also worth saying: my box has only an `i915` render node but *does* have `10_nvidia.json` installed, and your code correctly calls it single-GPU. That's precisely the case a naive implementation gets wrong, so nice.

Other things I confirmed: `bash -n` clean on all three scripts; `doctor` runs clean off the branch (exit 0, no unbound variables, no `set -e` casualties — and thank you for `count=$((count + 1))` rather than `((count++))`, which would have returned 1 on the first increment under `set -e`); and the duplicated installer block is byte-identical across the two files once comments are stripped, which given this repo's `ov02c10.yaml` history I did go and check.

**One blocking thing, and it's your own invariant.**

## Blocking: the global CPU debayer overrides the pin

`cmd_enable_persistent` makes the two exclusive — on a hybrid box `softisp_mode` stays empty, the unit gets the pin and no `LIBCAMERA_SOFTISP_MODE`. Correct as far as it goes.

But the installers' pre-existing block is untouched by this PR and still writes the mode *globally*:

```bash
if [[ -n "$NVIDIA_ACTIVE" ]]; then
    sudo mkdir -p /etc/environment.d
    echo "LIBCAMERA_SOFTISP_MODE=cpu" | \
        sudo tee /etc/environment.d/10-libcamera-softisp.conf > /dev/null
```

`/etc/environment.d` is consumed by `30-systemd-environment-d-generator`, which is a **user**-environment generator — so it lands in `camera-relay.service`. I measured both halves:

```
# ~/.config/environment.d/99-probe.conf -> PR76_PROBE_SOFTISP=cpu
$ systemctl --user start pr76-envprobe.service
inherited=cpu                 # environment.d does reach a user service

# + drop-in with Environment=PR76_PROBE_SOFTISP=UNIT_WINS
inherited=UNIT_WINS           # a unit-level Environment= does beat it
```

So on a hybrid machine where `NVIDIA_ACTIVE` is set, CPU mode arrives from the global file, wins (it touches no EGL at all), and the pin sits inert in the unit. Which is verbatim the thing your own comment says the design avoids:

> Deliberately exclusive. Setting both would leave CPU mode winning (it uses no EGL at all) and the pin dead in the unit on exactly the hybrid machines it was written for.

And this isn't a narrow edge. `NVIDIA_ACTIVE` is set when glxinfo reports NVIDIA, **or** when glxinfo is absent and `/proc/driver/nvidia` exists and `prime-select query != intel`. Neither installer installs `mesa-utils` — they only probe for it — and Ubuntu's default hybrid mode is `on-demand`, not `intel`. So *hybrid box with mesa-utils not installed* takes the fallback branch and gets the global CPU file. The message on that path already reads "couldn't confirm via glxinfo — install mesa-utils/glx-utils if your camera framerate is low", which is exactly the population you're trying to give the framerate back to.

It isn't a breakage — the camera works, on CPU debayer. It's a silent no-op of the feature on a real chunk of the target hardware, and `doctor` will print `vendor pin (service unit): …/50_mesa.json` while it's having no effect, which makes the new section actively misleading in the one scenario it exists for.

It also makes the `prime-select nvidia` gap you flagged worse than you realised: that configuration is *also* the one that gets the global CPU file, so as it stands it would get the pin and have it overridden.

Since a unit-level `Environment=` beats the inherited one, pin the mode alongside the vendor:

```diff
 local egl_vendor=""
 egl_vendor=$(detect_egl_vendor_pin) || egl_vendor=""
 if [[ -n "$egl_vendor" ]]; then
+    # Beat any global LIBCAMERA_SOFTISP_MODE=cpu from /etc/environment.d — it is
+    # inherited by this user service and would otherwise make the pin inert.
+    softisp_mode="gpu"
     info "Hybrid GPU detected — service will pin EGL to $egl_vendor (GPU debayer on the iGPU)"
```

`gpu` is a valid value — `strings libcamera.so.0.7` gives `", must be "cpu" or "gpu"`. Please leave the global file alone otherwise: it's what covers the PipeWire-libcamera path, which you correctly document as still unpinned.

## What needs to change for me to merge

1. **`softisp_mode="gpu"` in the hybrid branch of `cmd_enable_persistent`** (the diff above). Blocking.
2. **Print the effective debayer mode in `doctor`'s GPU / EGL section**, next to `vendor pin (service unit):`. The section reports the pin but not the mode, which is why the above is invisible to it — and it's the natural companion field. Blocking only in the sense that it's what stops this recurring silently.
3. **Put the "Keep this block in sync" comment on both installer copies** — `webcam-fix-book5/install.sh` has it, `webcam-fix-libcamera/install.sh` doesn't. Non-blocking but cheap, and this repo has been bitten by exactly that.
4. **A test for `detect_egl_vendor_pin`**, if you're willing — `camera-relay/tests/` exists now. The function has a clean seam: override `render_node_drivers` and the whole topology table falls out in a few lines. That locks the table in permanently rather than leaving it as a claim in a PR description. Happy to take the PR without it if you'd rather not.

Nit, take it or leave it: in `doctor`, `$(grep -oP '"library_path"…' "$vf" | head -1 || echo '?')` — the `||` binds to `head`, which exits 0 on empty input, so the `?` fallback never fires and a vendor JSON without a `library_path` renders as empty parens.

## What I couldn't check

No NVIDIA hardware here — this box is `i915`-only, so every hybrid result above is from mocked enumeration. The logic is confirmed; the behaviour on real hybrid silicon isn't, and rests on your Book4 Ultra evidence. I also can't test whether Mesa EGL on the iGPU render node works while NVIDIA is driving the display, which is the assumption the `prime-select nvidia` case stands on — so I'd still like a second tester there before we call that configuration solved, even after the fix above.

Fix 1 and 2 and I'll merge it. Thanks for the write-up, genuinely — the "it only breaks under systemd, so `cam` from a terminal lies to you" observation is the part that would have cost someone else a weekend.

4nrry:
All four addressed in `f5d6fed..a4ba63f`. You were right on the blocking one, and I verified the mechanism rather than taking it on trust — `30-systemd-environment-d-generator` is installed here, and the precedence half reproduces with a transient unit:

```
systemctl --user set-environment PR76_PROBE=from_manager
systemd-run --user … -- sh -c 'echo $PR76_PROBE'          -> from_manager
systemd-run --user -p Environment=PR76_PROBE=from_unit …  -> from_unit
```

So the pin was inert on any machine carrying the global cpu file, which — via the no-mesa-utils fallback branch you traced — is a large slice of the target hardware. My comment claiming the design avoided that was describing an invariant the code didn't actually hold.

**1. `softisp_mode="gpu"` in the hybrid branch** (`f5d6fed`). Confirmed `gpu` is accepted at runtime, not just present in the strings table — on the Book4 Ultra with the mode set explicitly: `GL_RENDERER: Mesa Intel(R) Arc(tm) Graphics (MTL)`, 30.08 fps. Global file untouched.

**2. Effective debayer mode in doctor** (`e082b2a`). Three layers, since only the effective one answers the question:

```
debayer mode (service unit): gpu
debayer mode (inherited):    cpu  (/etc/environment.d/10-libcamera-softisp.conf)
debayer mode (effective):    gpu (from the service unit)
```

plus an explicit `→ PIN IS INERT` block when a pin is present with no unit-level mode and an inherited cpu. Your nit is fixed in the same commit — you're right that the `||` bound to `head`, so the `?` never fired.

**3. Sync marker** (`636347f`).

**4. Tests** (`a4ba63f`) — `camera-relay/tests/test-egl-vendor-pin.sh`, 17 assertions, no hardware, no system state. Same seam you used: `find_mesa_egl_vendor` stubbed to a sentinel for the topology cases so they test the decision rather than the host's ICDs. Your twelve topologies plus a real-ICD content check, and a third group that generates an actual unit with every probe stubbed and asserts a hybrid box gets *both* the pin and `LIBCAMERA_SOFTISP_MODE=gpu`.

That last one is the guard for #1, and I checked it fails on the bug rather than only passing on the fix — reverting `softisp_mode="gpu"` drops the suite to 16/17 and exits 1.

Two things worth flagging:

- `test-gst-tools-check.sh` reports `10 passed, 1 failed` here — `doctor did not flag black frames`. It is **pre-existing and environment-dependent, not from this PR**: I ran the base-commit test against the base-commit script and got the identical failure. My relay is live and producing a real picture, so doctor correctly declines to flag black frames; the test looks like it assumes no working camera. Happy to open a separate issue if you want it tracked.
- The `prime-select nvidia` caveat still stands unchanged, and your framing is right that the fix above makes it *reachable* rather than solved. Mesa EGL on the iGPU render node while NVIDIA drives the display is still untested by either of us. I'd rather it stay called out in the PR body than quietly implied to work.

Andycodeman (approved):
All four verified here, and the way you handled the blocking one is the reason this was a pleasure to review. Thank you.

I checked each item rather than taking the reply on trust:

- **1 — `softisp_mode="gpu"`** (`f5d6fed`). Present, and the comment explaining *why* silence isn't enough is the part that stops someone helpfully "simplifying" it back out in a year.
- **2 — effective debayer mode** (`e082b2a`). The three-layer split is better than what I asked for — I only wanted the effective value; separating unit / inherited / effective means the next person can see *where* it came from, which is the whole diagnostic problem in miniature. The `→ PIN IS INERT` block is exactly right. Nit fixed in the same commit.
- **3 — sync marker** (`636347f`). One occurrence in each installer.
- **4 — tests** (`a4ba63f`). 18/18 here.

On the test suite specifically: you said it fails on the bug rather than only passing on the fix, and I checked that too — reverting `softisp_mode="gpu"` in a throwaway worktree:

```
  ✓ hybrid: unit pins the EGL vendor
  ✗ hybrid: unit does not state LIBCAMERA_SOFTISP_MODE=gpu — pin would be inert
  17 passed, 1 failed          SUITE EXIT=1
```

That's a real guard. The hygiene is good too — `mktemp -d` with a cleanup trap, `SERVICE_DIR` redirected, `systemctl` stubbed — and I confirmed my real `camera-relay.service` was byte-identical before and after the run before I trusted any of it. Stubbing `find_mesa_egl_vendor` to a sentinel for the topology cases, so those test the *decision* and not the host's ICDs, is the right call and is why this suite will still mean something on someone else's machine.

**Your `test-gst-tools-check.sh` note checks out.** I ran the same suite from a base-`main` worktree at `6c205ba`, without your PR, and got the identical `10 passed, 1 failed`. My relay is producing a real picture, so `doctor` correctly declines to flag black frames and the test's assumption doesn't hold. Pre-existing and environment-dependent, exactly as you said — please do open that issue, and thank you for checking it against the base commit instead of just asserting it.

Also re-confirmed at head: `bash -n` clean on all four scripts, and the duplicated installer block is still byte-identical across the two files.

**On `prime-select nvidia`** — I think you called that exactly right, and I want to endorse the judgement rather than just the outcome. Leaving it stated in the PR body instead of quietly implying it works is the correct move: the fix makes that path *reachable* rather than solved, and "Mesa EGL on the iGPU render node while NVIDIA drives the display" is still untested by either of us. It stays a known gap.

Worth knowing: my own box is a 960XGL Book4 Ultra with an RTX 4070 Max-Q, NVIDIA userspace installed and DKMS modules already built — but nothing bound to the device, so a single i915 render node and `EGL_VENDOR=Mesa Project` today. Which incidentally is a live confirmation of your mechanism from the other direction: the ICD sorts first and still loses, because it can't claim a device without a kernel driver. It also means I'm one module load from being the hybrid tester that caveat wants, so I may be able to close it myself.

Merging. Thanks again — you found a bug nobody had filed, took it to the mechanism, and then took the review correction further than it was written. That's the good kind of contribution.

Andycodeman:
Merged and shipped as **v0.3.57**. Thanks again.

One thing so you don't duplicate work: **I fixed the `test-gst-tools-check.sh` failure** rather than filing it — `aa0a195` on `main`. You were right that it's pre-existing and environment-dependent; the root cause turned out to be narrower than "the test assumes no working camera".

The test synthesises black frames by writing `videotestsrc` into the loopback with `v4l2sink`. But v4l2loopback takes a **single writer**, so on any machine where the relay is actually running, that `gst-launch` can never attach — it exits immediately, the real relay keeps feeding real frames, and `doctor` correctly reports `REAL PICTURE`, which the test scored as a doctor bug. Confirmed with `fuser`:

```
/dev/video0:  andy  3347  F....  camera-relay-mo
```

So it now checks it actually owns the writer before asserting, and skips otherwise — the same idiom the stale-PID case in that file already uses. `10 passed, 0 failed` here. I checked the guard both ways so it can't quietly swallow a real regression: live writer → assertion runs, dead writer → skips.

Also pushed `bb56528`, tightening something your PR surfaced indirectly. The docs said the EGL bug affects "every"/"all Galaxy Book4 Ultra" — but having the dGPU isn't sufficient, the NVIDIA driver has to be **bound** to it. My own 960XGL has an RTX 4070 Max-Q with NVIDIA userspace installed and DKMS built, yet nothing bound, so one render node and `EGL_VENDOR=Mesa Project`. `detect_egl_vendor_pin` already gets this right by enumerating render nodes; only the prose overstated it.

Looking at #77 now.

PR #77, opened 2026-08-04
docs: the 26 MHz OV02C10 clock is per-board, not Raptor Lake only

4nrry:
The docs attribute the `external clock 26000000 is not supported` probe failure exclusively to Raptor Lake IPU6. It happens on Meteor Lake too.

## Evidence

Galaxy Book4 Ultra **NP960XGL-XG1BR**, IPU6 **Meteor Lake `8086:7d19`** (not Raptor Lake `8086:a75d`), Kubuntu 26.04 / kernel 7.0.0-28-generic. Before the fix:

```
ov02c10 i2c-OVTI02C1:00: error -EINVAL: external clock 26000000 is not supported
ov02c10 i2c-OVTI02C1:00: probe with driver ov02c10 failed with error -22
```

After installing `ov02c10-26mhz-fix`, the sensor probes and the message becomes informational:

```
ov02c10 i2c-OVTI02C1:00: 26000000Hz clock (not 19200000Hz): sensor runs fast;
advertising pixel_rate=216666666 and padding vts to 3152 for ~30fps
```

Camera confirmed working end to end afterwards — `cam -c 1 -C5` captures at 30.08 fps, and the camera relay serves non-PipeWire apps.

## Why the wording matters

As written, a Meteor Lake owner reads "Raptor Lake" and concludes the fix doesn't apply to them — when it is the one thing standing between them and a working camera. I nearly skipped it myself for exactly that reason.

**The installer was already correct.** `webcam-fix-libcamera/install.sh` greps dmesg for the clock rejection and never inspects the platform, so this PR only corrects the prose to match what the code already does. No code changes.

## Changes

- `ov02c10-26mhz-fix/README.md` — drop "Raptor Lake" from the scope; state the clock is a per-board property; list confirmed models per platform; point readers at `dmesg` instead of the platform name; bump tested kernel ceiling 6.17 -> 7.0
- `webcam-fix-libcamera/README.md` — same correction in the troubleshooting entry
- `README.md` — same correction in the webcam fix overview; add the Book4 Ultra / Kubuntu 26.04 / kernel 7.0.0-28 entry to "Tested On"

## Note on the "Tested On" entry

All four DKMS modules (`ov02c10`, `ipu-bridge-fix`, `v4l2loopback`, `max98390-hda`) build unmodified against kernel 7.0 and load MOK-signed under Secure Boot. The internal mic needed no `mic-fix` there — kernel 7.0 already selects the SOF driver and exposes the DMIC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ The internal DMIC requires SOF (Sound Open Firmware) with a recent enough firmwa
 
 ### [Webcam Fix — Book3 / Book4](webcam-fix-libcamera/) — IPU6 + libcamera (Recommended)
 
-The built-in webcam uses Intel IPU6 (Meteor Lake or Raptor Lake) with an OmniVision OV02C10 sensor. This fix uses the open-source libcamera Simple pipeline handler with Software ISP, accessed through PipeWire. Includes IVSC module loading, initramfs configuration (eliminating the boot race condition), sensor tuning, WirePlumber rules to hide raw IPU6 nodes, and an on-demand camera relay for non-PipeWire apps (Zoom, OBS, VLC). The installer also auto-detects the [26 MHz clock issue](ov02c10-26mhz-fix/) affecting some Raptor Lake models and offers to install the DKMS fix.
+The built-in webcam uses Intel IPU6 (Meteor Lake or Raptor Lake) with an OmniVision OV02C10 sensor. This fix uses the open-source libcamera Simple pipeline handler with Software ISP, accessed through PipeWire. Includes IVSC module loading, initramfs configuration (eliminating the boot race condition), sensor tuning, WirePlumber rules to hide raw IPU6 nodes, and an on-demand camera relay for non-PipeWire apps (Zoom, OBS, VLC). The installer also auto-detects the [26 MHz clock issue](ov02c10-26mhz-fix/) — which affects individual boards on both Raptor Lake and Meteor Lake, so it is detected from the dmesg error rather than from the platform — and offers to install the DKMS fix.
 
 - PipeWire-native apps (Firefox, Chromium) access the camera directly — no relay needed
 - Non-PipeWire apps use the on-demand V4L2 relay: near-zero CPU when idle, camera activates only when an app opens the device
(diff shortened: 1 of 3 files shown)


Andycodeman (approved):
Merging this. You're right, and I can confirm it from the opposite direction — which turns out to make the point harder than the PR does.

My own machine is a Book4 Ultra **960XGL** on **Meteor Lake IPU6 `8086:7d19`** — same model designation and same platform as yours — and it does *not* have the 26 MHz clock:

```
00:05.0 Multimedia controller [8086:7d19]        # Meteor Lake IPU6
v4l-subdev6: ov02c10 3-0036                      # sensor probed fine
modinfo ov02c10 → /lib/modules/7.0.0-28-generic/ubuntu/ipu6/ov02c10.ko.zst
                                                 # stock in-tree driver, not DKMS
dkms status → max98390-hda only                  # no ov02c10 fix installed
journalctl -k -b | grep ov02c10 → "failed to check hwcfg: -517" and nothing else
```

That `-517` is `EPROBE_DEFER`, i.e. the normal deferred-probe retry, not a failure. Camera runs end to end at 30 fps on the unpatched driver. I checked `modinfo` and `dkms status` specifically because if I'd had your fix applied I'd have drawn the opposite conclusion — I don't.

So two Book4 Ultra 960XGL boards, both Meteor Lake IPU6 `8086:7d19`, disagree. The clock isn't merely platform-independent; it isn't even **model**-determined. Your "go by the dmesg error" advice is the only reliable rule, and it's now better supported than when you wrote it.

Your supporting claims check out too:

- **"The installer was already correct."** Confirmed — `webcam-fix-libcamera/install.sh:300-306` greps `dmesg` / `sudo dmesg` / `journalctl -k` for the clock rejection and never inspects the platform. This really is prose catching up to code.
- **Kernel ceiling 6.17 → 7.0.** Fine — I'm on 7.0.0-28-generic with the in-tree `ov02c10` loaded and working.
- **"Kernel 7.0 selects SOF and exposes the DMIC, no mic-fix needed."** Reproduced on the same kernel here: `snd_sof_pci_intel_mtl` loaded, `card 0 … device 6: DMIC Raw` present, nothing mic-related in `/etc/modprobe.d/`.

## One change I'm making on top, rather than sending back to you

The affected-models list carries the bare model:

> - Galaxy Book4 Ultra **NP960XGL** — Meteor Lake IPU6 `8086:7d19`

Read literally that tells every 960XGL owner they need the DKMS patch — including mine, which demonstrably doesn't. That's the same failure this PR fixes, just inverted: instead of Meteor Lake owners wrongly skipping the fix, 960XGL owners would wrongly install it.

Your board is **NP960XGL-XG1BR**; the SKU suffix is doing real work. So I'll follow up with a commit that carries the full SKU in the affected list and states outright that the same model can differ board to board, citing the 960XGL that doesn't need it. I'm doing it myself rather than requesting changes because the counter-evidence is my hardware and you have no way to produce it — no reason to cost you a round trip for that.

It strengthens your argument rather than qualifying it: "not platform-determined" becomes "not even model-determined, check dmesg."

Thanks for this. "I nearly skipped it myself for exactly that reason" is the most useful sentence in the PR — that's the actual cost of the old wording, and it's why this was worth fixing rather than living with.

Andycodeman:
Follow-up pushed as `1f30ca6` — the SKU change I flagged above, so it's done rather than sitting on a TODO.

- affected list now carries **NP960XGL-XG1BR** in all three files
- added the explicit "not even model-determined" note, citing the 960XGL that runs the stock driver clean
- reframed the list as *boards seen affected*, never *models*

Your framing survives intact — it just got one step stronger: don't decide by platform **or** by model, go by the dmesg error. Which is what `install.sh` was doing all along.

Thanks again.

PR #78, opened 2026-08-04
webcam-fix-libcamera: actually hide the raw MC-centric camera nodes

4nrry:
The installer promises that the ~48 raw IPU6 nodes stay out of application camera lists. They don't — here they are in Zen (Firefox-based), and they show up in Zoom, OBS and colord too.

## Why the current rule can't work

`90-hide-ipu6-v4l2.rules` fails for two independent reasons, either of which is enough on its own.

**It targets the wrong thing.** The rule only does `TAG-="uaccess"`, but [`install.sh:1332`](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/blob/main/webcam-fix-libcamera/install.sh#L1332) adds the user to `video` and the nodes are `root:video 0660`. The group is what grants access, so removing the session ACL tag changes nothing. On my machine `/dev/video1` had no ACL and `open()` still succeeded.

**`TAG-="uaccess"` can't stop the ACL anyway.** `70-uaccess.rules` tags every `video4linux` device unconditionally, and the `TAG==` test in `73-seat-late.rules` matches the *sticky* tag list, which `TAG-=` doesn't touch. `udevadm test` shows it plainly:

```
CURRENT_TAGS=:uaccess:seat:...      ← TAG-= did remove it here
TAGS=:uaccess:...                   ← but 73 matches this one
Queued commands:
  RUN{builtin} : uaccess
```

No rule priority fixes that, and `70-uaccess.rules` has no usable early exit.

## The kernel already tells us

The ISYS driver sets **`V4L2_CAP_IO_MC`** on all 48 nodes — they only produce frames once userspace configures the media graph, and are never usable as standalone cameras. The loopback doesn't have the bit:

```
/dev/video1  Device Caps 0x24a00001   (I/O MC)
/dev/video0  Device Caps 0x05200003   (Camera Relay)
```

Nothing carries that into udev: systemd's [`v4l_id`](https://github.com/systemd/systemd/blob/main/src/udev/v4l_id/v4l_id.c) derives `ID_V4L_CAPABILITIES` from `V4L2_CAP_VIDEO_CAPTURE` alone and never looks at `IO_MC`. So all 48 claim to be cameras.

## Two kinds of application, two answers

- **Chromium and friends** enumerate through udev — clearing the properties is enough.
- **Firefox/libwebrtc** ([`device_info_v4l2.cc`](https://searchfox.org/mozilla-central/source/third_party/libwebrtc/modules/video_capture/linux/device_info_v4l2.cc)) walks `/dev/video0..63` calling `QUERYCAP`, ignores udev entirely, and only skips a node whose `open()` fails. **Permissions are the only lever there.**

Renaming the nodes out of `/dev/video*` isn't available: udev(7) is explicit that a device node name cannot be changed, only symlinked.

## What this does

- **`v4l2-io-mc.c`** — udev helper exporting the kernel's capability as `ID_V4L_IO_MC`, so the rule keys on that instead of a driver-specific card name. Covers IPU6, IPU7 and anything else media-controller based.
- **`74-camera-relay-mc-nodes.rules`** — moves the nodes to a memberless `camera-relay` group, clears the properties that make them look like cameras (including `ID_V4L_PRODUCT`, since `95-cd-devices.rules` keys on that rather than the capabilities), and strips the ACL with a `RUN` that lands after 73's builtin. The `74-` prefix is load-bearing; the file explains why.
- **`camera-relay-gst.c`** — setgid launcher giving the libcamera pipeline that group. The UID is unchanged, which is what keeps the EGL context for GPU debayer, the uaccess ACLs on the loopback and `/dev/media0`, and the monitor's same-UID `/proc` scan all working. It assembles the pipeline itself rather than running a command line it was handed, and builds a fresh environment instead of filtering the inherited one, since GStreamer and libcamera both load code from paths named by environment variables.

`detect_camera_name` now goes through the launcher as well — `cam -l` is the method actually in use (the PipeWire libcamera node is disabled on purpose) and it can no longer open the nodes as the user. `camera-relay doctor` gained checks for the setgid bit and for any MC node still reachable from the session; both fail silently otherwise. The `--foreground` path picks up `RELAY_COLOR_FILTER` as a side effect — it used to drop it.

## Interaction with #76 (second commit)

Building a fresh environment also drops `__EGL_VENDOR_LIBRARY_FILENAMES`, which is exactly what #76 added to keep the debayer off NVIDIA's EGL driver. Rebasing onto #76 alone was not enough — the launcher has to forward the pin explicitly, or every frame comes out black with #76's own signature:

```
ERROR eGL egl.cpp:134 glFrameBufferTexture2D error 36054
ERROR Debayer debayer_egl.cpp:639 debayerGPU failed
```

So `--egl-vendor` is forwarded and validated, against a narrower prefix list than the other paths (`/etc/glvnd/egl_vendor.d`, `/usr/share/glvnd/egl_vendor.d`) because the value picks which driver runs the debayer. That commit also closes a hole in `--list-cameras`, which exported `--ipa-path`/`--gst-plugin-path` into the child environment without validating them.

## Verification

On a Book4 Ultra (i915 + nvidia), after a **cold boot**:

```
/dev/video1  root camera-relay crw-rw----    ← no ACL
/dev/video0  root video        crw-rw----+   ← untouched
open() video1: denied    open() video0: allowed
leaking: 0 of 48
doctor: Launcher setgid OK | MC nodes hidden OK | EGL_VENDOR: Mesa Project | REAL PICTURE
```

The picker lists only `Camera Relay`:

Camera picker showing a single entry, Camera Relay

`tests/test-launcher-validation.sh` covers the launcher's input validation — 26 cases, hardware-free, no system state touched. The existing suite is unchanged at 10/11 with the same pre-existing failure (verified against the pristine script via `git stash`).

## Notes for review

Two things deliberately left out, and one thing worth a closer look.

`webcam-fix-book5` (IPU7) still uses a name-matched rule and would be covered by the same generic `IO_MC` rule — left out because I have no Book5 to validate on.

`install.sh` and `uninstall.sh` are syntax-checked but were never run end to end: this was applied to a live machine by installing the individual pieces in a safe order, not by re-running the installer. The commands themselves are all exercised, but the full installer path is untested.

The setgid binary is the security-sensitive part and deserves a maintainer's eye. The threat model is narrow — the group grants only raw camera access, which the user already reaches through the relay — but it is a permanent fixture, so the two properties it exists to hold (never exec a caller-supplied command line; build a fresh environment rather than filtering the inherited one) are worth checking against `tests/test-launcher-validation.sh`.

### DIFF
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -24,6 +24,24 @@ SERVICE_DIR="${HOME}/.config/systemd/user"
 SERVICE_NAME="camera-relay.service"
 LOOPBACK_CONF="/etc/modprobe.d/99-camera-relay-loopback.conf"
 MONITOR_BIN="/usr/local/bin/camera-relay-monitor"
+LAUNCHER_BIN="/usr/local/bin/camera-relay-gst"
+
+# Arguments shared by every launcher invocation. The launcher is setgid, so it
+# replaces its environment wholesale rather than trusting what it inherited —
+# the paths setup_environment exported have to be handed over as arguments.
+LAUNCHER_ARGS=()
+build_launcher_args() {
+    LAUNCHER_ARGS=()
+    [[ -n "${GST_PLUGIN_PATH:-}" ]] && \
+        LAUNCHER_ARGS+=(--gst-plugin-path "$GST_PLUGIN_PATH")
+    [[ -n "${LIBCAMERA_IPA_MODULE_PATH:-}" ]] && \
+        LAUNCHER_ARGS+=(--ipa-path "$LIBCAMERA_IPA_MODULE_PATH")
+    [[ -n "${LIBCAMERA_SOFTISP_MODE:-}" ]] && \
+        LAUNCHER_ARGS+=(--softisp-mode "$LIBCAMERA_SOFTISP_MODE")
+    [[ -n "${RELAY_COLOR_FILTER:-}" ]] && \
+        LAUNCHER_ARGS+=(--color-filter "$RELAY_COLOR_FILTER")
+    return 0
+}
 
 # ── Helpers ──────────────────────────────────────────────────────────────────
 
@@ -63,23 +81,29 @@ detect_camera_name() {
     # Output format: "1: 'ov02c10' (\_SB_.PC00.LNK0)"
     # Extract the camera ID from parentheses — this is what gst-launch needs
     # for the camera-name property (not the display name in quotes).
-    local cam_cmd=""
-    # Prefer /usr/local/bin/cam only if it actually runs (not broken by stale libs)
-    if [[ -x /usr/local/bin/cam ]] && /usr/local/bin/cam -l &>/dev/null; then
-        cam_cmd="/usr/local/bin/cam"
+    # The raw camera nodes belong to the memberless camera-relay group, so a
+    # plain `cam` run as the user can no longer enumerate them. Go through the
+    # launcher, which holds the group. Direct `cam` remains the fallback for
+    # installs without the launcher and for hardware whose nodes were never
+    # restricted (nothing MC-centric to hide).
+    local -a cam_cmd=()
+    if [[ -x "$LAUNCHER_BIN" ]] && "$LAUNCHER_BIN" --list-cameras &>/dev/null; then
+        cam_cmd=("$LAUNCHER_BIN" --list-cameras)
+    elif [[ -x /usr/local/bin/cam ]] && /usr/local/bin/cam -l &>/dev/null; then
+        cam_cmd=(/usr/local/bin/cam -l)
     elif command -v cam &>/dev/null; then
-        cam_cmd="cam"
+        cam_cmd=(cam -l)
     fi
-    if [[ -n "$cam_cmd" ]]; then
+    if (( ${#cam_cmd[@]} > 0 )); then
         local name
-        name=$("$cam_cmd" -l 2>/dev/null | grep -oP '\(([^)]+)\)' | head -1 | tr -d '()')
+        name=$("${cam_cmd[@]}" 2>/dev/null | grep -oP '\(([^)]+)\)' | head -1 | tr -d '()')
         if [[ -n "$name" ]]; then
             echo "$name" > "$CAMERA_CACHE"
             echo "$name"
             return 0
         fi
         # Fallback to quoted display name if no parenthesized ID
-        name=$("$cam_cmd" -l 2>/dev/null | grep -oP "'[^']+'" | head -1 | tr -d "'")
+        name=$("${cam_cmd[@]}" 2>/dev/null | grep -oP "'[^']+'" | head -1 | tr -d "'")
         if [[ -n "$name" ]]; then
             echo "$name" > "$CAMERA_CACHE"
             echo "$name"
@@ -558,30 +582,20 @@ start_pipeline() {
     local fast="${3:-false}"
     local gst_log="${XDG_RUNTIME_DIR:-/tmp}/camera-relay-gst.log"
 
-    # gst-launch-1.0 interprets backslashes as escape characters.
-    # Camera names like "\_SB_.LNK0" need double-escaping.
-    local gst_camera_name="${camera_name//\\/\\\\}"
-
-    # Use io-mode=mmap for v4l2sink — kernel-managed buffers are the most
-    # compatible across v4l2loopback versions. io-mode=rw (1) breaks on
-    # some Fedora systems; io-mode=auto (0) may fall through to rw.
-    # GStreamer enum: 0=auto, 1=rw, 2=mmap, 3=userptr, 4=dmabuf
-    # RELAY_COLOR_FILTER: optional GStreamer element(s) for color correction.
-    # Set in service environment to apply per-sensor tuning, e.g.:
+    # The launcher assembles the pipeline itself — it is setgid and must never
+    # run a command line handed to it. Backslash escaping for the gst-launch
+    # parser (camera names look like "\_SB_.LNK0") happens in there too.
+    # RELAY_COLOR_FILTER: optional GStreamer element(s) for color correction,
+    # set in the service environment for per-sensor tuning, e.g.
     #   RELAY_COLOR_FILTER="videobalance saturation=0.85"
-    local -a color_filter=()
-    if [[ -n "${RELAY_COLOR_FILTER:-}" ]]; then
-        color_filter=(! $RELAY_COLOR_FILTER)
-    fi
+    # The launcher accepts only allow-listed video filters.
+    build_launcher_args
 
     local -a gst_cmd=(
-        gst-launch-1.0 -e
-        libcamerasrc camera-name="$gst_camera_name"
-        ! queue max-size-buffers=3 leaky=downstream
-        ! videoconvert
-        "${color_filter[@]}"
-        ! "video/x-raw,format=YUY2"
-        ! v4l2sink device="$loopback_dev" io-mode=mmap sync=false
+        "$LAUNCHER_BIN"
+        --camera "$camera_name"
+        --v4l2-sink "$loopback_dev"
+        "${LAUNCHER_ARGS[@]}"
     )
 
     "${gst_cmd[@]}" >"$gst_log" 2>&1 &
@@ -666,13 +680,13 @@ cmd_start() {
         info "Starting relay (foreground)..."
         echo "streaming" > "$STATE_CACHE"
         echo $$ > "$PID_FILE"
-        local gst_camera_name="${camera_name//\\/\\\\}"
-        exec gst-launch-1.0 -e \
-            libcamerasrc camera-name="$gst_camera_name" \
-            ! queue max-size-buffers=3 leaky=downstream \
-            ! videoconvert \
-            ! "video/x-raw,format=YUY2" \
-            ! v4l2sink device="$loopback_dev" io-mode=mmap sync=false
+        # Same launcher as the backgrounded path, so this one now honours
+        # RELAY_COLOR_FILTER too — it used to silently drop it.
+        build_launcher_args
+        exec "$LAUNCHER_BIN" \
+            --camera "$camera_name" \
+            --v4l2-sink "$loopback_dev" \
+            "${LAUNCHER_ARGS[@]}"
     else
         info "Starting relay..."
         local pid
@@ -732,6 +746,13 @@ cmd_start_on_demand() {
 Run the installer to build it, or use 'camera-relay start' for always-on mode."
     fi
 
+    # Without the launcher the pipeline cannot open the raw camera nodes —
+    # they belong to the camera-relay group, which this session is not in.
+    if [[ ! -x "$LAUNCHER_BIN" ]]; then
+        die "camera-relay-gst not found at $LAUNCHER_BIN.
+Run the installer to build it — the pipeline needs it to reach the camera."
+    fi
+
     echo $$ > "$PID_FILE"
     echo "idle" > "$STATE_CACHE"
 
@@ -745,20 +766,14 @@ Run the installer to build it, or use 'camera-relay start' for always-on mode."
     # YUY2 frames from its stdout, and relays them to the device.
     # This avoids the writer fd handoff gap — the monitor always holds
     # the device open, so ready_for_capture never drops.
-    local gst_camera_name="${camera_name//\\/\\\\}"
-    local -a color_filter=()
-    if [[ -n "${RELAY_COLOR_FILTER:-}" ]]; then
-        color_filter=(! $RELAY_COLOR_FILTER)
-    fi
+    build_launcher_args
 
     local -a gst_cmd=(
-        gst-launch-1.0 -e
-        libcamerasrc camera-name="$gst_camera_name"
-        ! queue max-size-buffers=3 leaky=downstream
-        ! videoconvert
-        "${color_filter[@]}"
-        ! "video/x-raw,format=YUY2,width=1920,height=1080"
-        ! fdsink fd=3 sync=false
+        "$LAUNCHER_BIN"
+        --camera "$camera_name"
+        --fd-sink 3
+        --width 1920 --height 1080
+        "${LAUNCHER_ARGS[@]}"
     )
 
     # Clean up all children on exit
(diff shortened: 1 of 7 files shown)


Andycodeman (changes requested):
You asked for a maintainer's eye on the setgid binary. There's a real bug in it, so I'll lead with that — but I want to be clear up front that the analysis behind this PR is excellent. Tracing the failure to `73-seat-late.rules` matching the *sticky* tag list that `TAG-=` doesn't touch, and demonstrating it with `udevadm test` rather than asserting it, is the kind of root-causing that makes a rule change reviewable at all. Same for reading `v4l_id.c` to establish that nothing carries `V4L2_CAP_IO_MC` into udev, and for splitting udev-enumerating apps from libwebrtc's `open()`-probing walk — that's the distinction the whole design rests on and it's correct.

## Blocking: stack buffer overflow in the setgid binary

`camera-relay-gst.c:476` — `gst_argv[n] = NULL;` writes one pointer past the end of `gst_argv[MAX_ARGS]`.

```
==429872==ERROR: AddressSanitizer: stack-buffer-overflow
WRITE of size 8 at 0x7f8a4ac00270 thread T0
    #0 ... in main camera-relay-gst.c:476
Address ... is located in stack of thread T0 at offset 624 in frame
    [112, 624) 'gst_argv' (line 339) <== Memory access at offset 624 overflows this variable
```

`append_color_filter` guards with `argc >= MAX_ARGS - 8`, reserving 8 slots. The tail after it needs **nine**: `!`, caps, `!`, `v4l2sink`, `device=…`, `io-mode=mmap`, `sync=false` — seven — plus the `NULL` terminator, and the element branch can push `argc` two past the guard in a single iteration. So on the `--v4l2-sink` path, which is one element longer than `--fd-sink`, index 64 gets written on a 64-element array.

Reachable straight from argv, which for a setgid binary is attacker-controlled — 15 chained filters does it. Exact boundary, built `-fsanitize=address,undefined`:

```
k=13  accepted, no overflow
k=14  accepted, no overflow
k=15  *** ASAN: stack-buffer-overflow ***
k=16  rejected (guard)
```

with `--camera testcam --v4l2-sink /dev/video9 --color-filter "videoconvert ! videoconvert ! …"`.

It writes `NULL` rather than attacker data, so I'm not claiming a clean exploit — but an out-of-bounds write in a permanently-installed setgid binary isn't something to reason about case by case.

**One character fixes it.** `MAX_ARGS - 8` → `MAX_ARGS - 9`, verified across the same sweep:

```
k=14  accepted, no overflow
k=15  rejected (guard)      ← previously the overflow
k=20  rejected (guard)
```

I'd also add a comment doing the arithmetic explicitly, since "8" vs "9" is exactly the sort of constant that drifts when the tail gains an element.

## Your test suite passes and doesn't catch it

`tests/test-launcher-validation.sh` is **26/26 green** here, including with the overflow present. That's not a criticism of the cases — they're well chosen — but they all test *input rejection*, and the suite compiles without sanitizers. For a setgid binary the memory-safety property is the one that matters most, and right now nothing exercises it.

Two additions would close that:

- build the binary under `-fsanitize=address,undefined` for the test run
- add a maximal-length `--color-filter` case on **both** sink paths, since `--v4l2-sink` is the longer one and the shorter path masked this

That turns a bug I happened to count into one the suite catches.

## Non-blocking

**Stale rules filename in the header comment.** `camera-relay-gst.c:6` says:

> see `90-camera-relay-mc-nodes.rules`

but the shipped file is `74-camera-relay-mc-nodes.rules` (`install.sh:1513`), and both `install.sh` and `uninstall.sh` clean up old `90-` and `72-` names — so it looks like a rename that the comment missed. Normally cosmetic; here it isn't, because your own description says **"The `74-` prefix is load-bearing"**. Pointing a future maintainer at `90-` in the one place ordering is critical is worth fixing.

**`--v4l2-sink` length is unbounded.** `valid_ident(v4l2_sink + 10, "")` restricts the characters but not the length, so `/dev/video` plus 100 digits validates and then truncates into `sink_arg[64]`. `snprintf` makes that safe, and the result just fails to open — but a length check would say what you mean.

## What I'm not signing off on

Your own two caveats stand and I'd repeat them back rather than let them get lost:

- **`install.sh`/`uninstall.sh` were never run end to end.** You applied the pieces by hand in a safe order. That's the honest thing to say, and it's also the part I can't check for you — a udev rule plus a new group plus a setgid binary is exactly the combination where installer ordering bites, and a half-applied state could leave the raw nodes group-owned with no launcher to reach them.
- **Book5/IPU7 left out** for want of hardware. Agreed that's the right call over guessing.

I also haven't verified the security design end to end on hardware — I read it and tested the launcher in isolation. On that reading the two properties you set out to hold do hold: it assembles the pipeline itself rather than exec'ing anything it was handed, it replaces the environment wholesale rather than filtering it, `require_prefix` rejects `..` so the prefix check isn't defeated by traversal, and the `setresgid` rationale for avoiding `AT_SECURE` is sound and well-explained. The narrower prefix list for `--egl-vendor` than for plugin paths is a nice touch — that value picks which driver runs the debayer, and it deserves the tighter treatment.

Fix the overflow and get the suite building with sanitizers, and I'll take another pass. This is good work; it just needs the memory-safety property nailed down before a setgid binary lands on other people's machines.

4nrry:
Overflow confirmed and fixed in 0a57a9c. I reproduced it before touching the constant — my boundary landed at k=16 rather than your k=15, just a difference in how we count chained filters, same bug:

```
k=14  accepted, no overflow
k=15  accepted, no overflow
k=16  *** OVERFLOW ***
k=17  rejected (guard)
```

Rather than bump `8` to `9`, the reserve is now a named `TAIL_SLOTS` with the arithmetic written out beside it — your point about that constant drifting is exactly right, and the derivation is the part worth keeping:

```
"!" caps "!" v4l2sink device=… io-mode=mmap sync=false   → 7
NULL terminator                                          → 8
one loop iteration can append two entries at once        → 9
```

**On the suite.** You're right that this was the wrong property to have covered, and the failure mode is worse than "a case was missing" — 26/26 green with an out-of-bounds write present is a suite that was actively misleading about the thing that matters most for a setgid binary. It now builds a second binary under `-fsanitize=address,undefined` and runs every case against it, plus a sweep of 1..24 chained filters on **both** sink paths.

I checked that the sweep actually catches this rather than assuming it — restoring `TAIL_SLOTS` to 8:

```
✗ --v4l2-sink with 16 filters — sanitizer report
    camera-relay-gst.c:496: runtime error: index 64 out of bounds for type 'char *[64]'
  exit code: 1
```

Both non-blocking items are in too: the header comment now points at `74-`, and `--v4l2-sink` bounds the device number's length instead of leaning on `snprintf` to truncate safely.

**On what you're not signing off.** Both caveats stand, and the installer one is the one I'd flag hardest: still applied by hand in a safe order, never run end to end. Your framing of the risk is sharper than mine was — a half-applied state leaving the raw nodes group-owned with no launcher to reach them is a dead camera, not a degraded one, and the installer writes the rule before it builds the launcher. If you'd rather I reorder those two steps so the launcher exists before the nodes change hands, say the word; it's a small change and it removes the bad intermediate state entirely.

Suites here: launcher 29/29 (sanitized), gst-tools 10/10, egl-vendor-pin 17/17.

4nrry:
Closing out the installer caveat, and one fix that came out of trying.

**The ordering you flagged was real, and worse than "a failure in between".** The rule was written at ~1513 and triggered at ~1572; the launcher was not built until ~1887. That is ~315 lines during which the raw nodes already belonged to `camera-relay` and nothing existed that could open them. And the launcher build sat inside the `[[ -d "$RELAY_DIR" ]]` block, so a tree without `camera-relay/` would install the rule and never build a launcher at all — not a window, a permanent dead camera.

8b7dc02 builds both helpers up front and installs the rule only if both succeeded. If either fails the nodes keep their old ownership: spurious entries still listed, camera still working. Degraded list beats unopenable device.

**`install.sh` has now been run end to end** on the Book4 Ultra, from a clean worktree on this branch:

```
rules:    70-camera-relay-capabilities  74-camera-relay-mc-nodes
video1:   root:camera-relay 660   ACL: 0 entries
video0:   root:video 660          (untouched)
launcher: root:camera-relay 2755
MC nodes reachable: 0 of 48
doctor:   Launcher setgid OK | MC nodes hidden OK | REAL PICTURE
```

One thing the run surfaced: `enable-persistent` reports "already enabled" and returns without regenerating anything, so the installer swaps binaries underneath a running relay without restarting it. Harmless here — the launcher is `exec`d per pipeline, not resident — but it means the install alone doesn't demonstrate the new set survives a restart. I restarted explicitly; `REAL PICTURE` after.

What this run does **not** cover: it started from an already-installed system, so it exercised the reinstall path. The ordering bug only bites on a clean install. I can test that path too — remove the group, rule and launcher first — but it opens a real window with no camera on my only machine, so I'd rather do it deliberately than fold it into this.

`uninstall.sh` remains unrun.

Andycodeman:
Re-checked everything on `8b7dc02` and all four items hold up. Verified independently rather than from the summaries:

- **Overflow closed.** My own ASan sweep, 1–24 chained filters, on **both** sink paths: clean throughout. `TAIL_SLOTS` with the derivation spelled out beside it is better than the bare constant I suggested — that's the part that stops it drifting when the tail changes.
- **The suite is now a real guard.** 29/29 sanitized, and I mutation-tested it: `TAIL_SLOTS` back to 8 gives `28 passed, 1 failed`, exit 1. It catches the bug rather than merely passing on the fix.
- **Header comment** points at `74-`, and `--v4l2-sink` bounds the device number (`✓ an over-long device number`).
- **Installer ordering.** Confirmed at the source: both helpers built at 1517–1534, rule written at 1540 only under `if [[ -n "$MC_HELPER_OK" && -n "$LAUNCHER_OK" ]]`. A failed build now leaves the nodes with their old ownership — spurious entries, working camera. That's the right trade and it removes the bad intermediate state entirely.

Your read on the ordering risk was sharper than my original note, too: I described a window, and you found it was also *permanent* for any tree without `camera-relay/`, because the build sat inside the `[[ -d "$RELAY_DIR" ]]` block. That's the worse half and I'd missed it.

**One thing before I merge: `uninstall.sh` has the same bug, mirror-imaged.**

You fixed the ordering in `install.sh`. The uninstaller still has it in reverse:

```
line  31:  sudo rm -f /usr/local/bin/camera-relay-gst        ← launcher gone
   …23 privileged commands, under set -e…
line 121:  sudo rm -f …/74-camera-relay-mc-nodes.rules       ← rule still live until here
line 130:  sudo udevadm trigger …                            ← ownership only reverts here
```

For ~90 lines the nodes still belong to `camera-relay` and nothing exists that can open them — the exact dead-camera state you just eliminated from the installer. `set -e` at line 6 means any failure in that window makes it permanent, and `uninstall.sh` is precisely the script someone reaches for to *recover* from a broken state, so it's the worst place to leave it.

The fix mirrors what you already did: drop the rule and trigger udev **first**, so ownership has reverted before the launcher is removed. Then the failure mode is "spurious nodes are back, camera works", which is where you want to land.

I'd rather you made that change than me. It's three lines, but neither of us has run this script and you have the hardware to actually exercise it — unlike the #77 correction, where I was holding evidence you couldn't get.

Your remaining caveats I'm content to merge alongside, since you've stated them rather than papered over them: the clean-install path is untested (and is now the safe-by-construction one), `uninstall.sh` unrun — which is what turned this up — and Book5/IPU7 out of scope for want of hardware. Also noted that `enable-persistent` doesn't regenerate on reinstall; worth its own issue rather than growing this PR.

For sequencing: **#79 is merged** (`a6372a8`). It touched the same function, and `git merge-tree` agreed with you that they don't conflict, but you'll want to rebase before this lands so the merged result is what actually gets tested.

Reorder the uninstaller and this is good to go.

Andycodeman (approved):
Merging. The reorder is exactly right, and I re-verified against the rebased head rather than the earlier results, since every SHA moved:

**Uninstall ordering.** Step 0 now drops the rule at line 29 and triggers udev at 34, with the launcher not removed until 50 — ownership has reverted before anything can be left unopenable. You also stripped the now-duplicate removals out of `[4/8]`, so the `74-` rule appears exactly once. It sits after the `EUID` guard, and everything in it is failure-safe under `set -e`: `rm -f` returns 0 on a missing file and both `udevadm` calls carry `|| true`. Failure anywhere below now lands on "spurious nodes are back and the camera works", which is where I wanted it.

The commit message is the part I'd single out — it records *why* the uninstaller is the worst place for that window, not just that the order changed. That's what stops someone tidying it back later.

**Re-verified on `51f4a31`:**

| check | result |
| --- | --- |
| `bash -n` — both installers, relay, all five test scripts | clean |
| independent ASan sweep, 1–24 chained filters, **both** sink paths | clean |
| `test-launcher-validation.sh` (sanitized) | 29/29 |
| `test-monitor-exit-propagation.sh` | 5/5 |
| `test-egl-vendor-pin.sh` | 18/18 |
| `test-writer-format-check.sh` | 1/1 (skips with a live relay, as designed) |

Also confirmed the branch is a genuine descendant of current `origin/main`, so those numbers are the merged result with #79 in, not the pre-rebase branch. Good call rebasing before it landed.

**On the remaining gaps** — I'm merging with them open, on purpose, and I want that on the record rather than implied:

- **`uninstall.sh` still has never been run.** I'm explicitly *not* asking you to run it. It's your only machine, running it removes your working camera fix, and the reorder is precisely what makes an untested run tolerable — the failure mode is now degraded, not dead. That trade is the right one.
- **Clean-install path untested**, but it's now the safe-by-construction one: the rule is written only when both helpers built.
- **Book5/IPU7 out of scope** for want of hardware. Right call over guessing.
- **`enable-persistent` doesn't regenerate on reinstall** — please do open that as its own issue rather than growing this PR.

This one was worth the rounds. You shipped a setgid binary with an out-of-bounds argv write, a test suite that was green while it was present, and an installer that could leave the nodes unopenable — and by the end all three are fixed, the suite catches its own regression under ASan, and the failure modes on both install and uninstall degrade rather than brick the camera. The `TAIL_SLOTS` derivation written out beside the constant is better than the bare `9` I suggested, and finding that the launcher build sat inside `[[ -d "$RELAY_DIR" ]]` — permanent, not a window — was the half I'd missed.

Cutting a release with this and #79 in it.

PR #79, opened 2026-08-04
camera-relay-monitor: refuse to stream into a format the device didn't take

4nrry:
`open_writer()` asks the loopback for 1920x1080 YUYV and then carries on regardless of the answer:

```c
if (xioctl(fd, VIDIOC_S_FMT, &fmt) < 0)
        fprintf(stderr, "[monitor] S_FMT warning: %s\n", strerror(errno));
```

Two ways that goes wrong. The ioctl can fail outright — and it can *succeed after adjusting what it was given*, since V4L2 reports the result back in the same struct, so a zero return is not confirmation.

Either way the monitor then writes fixed-size YUYV frames into a device on some other format, which is garbage to every reader. Nothing surfaces it: the device stays open, `camera-relay status` still says `STREAMING`, apps show a black picture, and the one warning lands in a log the next restart truncates.

## Why this is reachable

The loopback advertises only the writer's format while a writer holds it. With no writer it offers everything it can carry, and any reader may select one. On a Book4:

```
$ v4l2-ctl -d /dev/video0 --list-formats     # relay writing
        [0]: 'YUYV' (YUYV 4:2:2)

$ systemctl --user stop camera-relay.service
$ v4l2-ctl -d /dev/video0 --list-formats     # no writer
        [0]: 'BGR4' … [41]: 'MJPG' … [57]: 'HEVC'     # 58 of them

$ v4l2-ctl -d /dev/video0 --set-fmt-video=pixelformat=MJPG
$ v4l2-ctl -d /dev/video0 --get-fmt-video
        Pixel Format      : 'MJPG' (Motion-JPEG)
```

So a client that outlives the relay can leave the device somewhere else.

To be straight about the provenance: I hit a black-picture incident that looked exactly like this and could **not** reproduce it afterwards — the relay reclaims YUYV even with a reader attached, streaming or not. So this is not a fix for a confirmed field failure. It is a silent-failure path found while chasing one, and the cost of hitting it is an hour of debugging with no signal pointing anywhere.

## The change

`open_writer()` now fails on both, naming what the device is actually on and why it stopped:

```
[monitor] /dev/video0 stayed on 8192x1080 YUYV after asking for 9000x1080 YUYV.
[monitor] Fixed-size YUYV frames written to it would be garbage to every
[monitor] reader, so the relay stops here rather than stream silently. If an
[monitor] application is holding the device with its own format, closing it
[monitor] lets the relay recover on the next start.
```

Both call sites already handle a negative return, and the unit carries `Restart=on-failure` with `RestartSec=5` — so the relay retries and recovers on its own once whatever held the device lets go, rather than sitting there streaming into the void.

## Tests

`tests/test-writer-format-check.sh`. Triggering a mismatch needs only a geometry past the device's `max_width`, which produces the same shape of disagreement as a foreign format without having to wrestle a second client onto the device. It skips when no idle loopback is available and never touches one a running relay is driving — verified both ways:

```
with the relay running:   – skipped: no idle v4l2loopback device to test against
with it stopped:          4 passed, 0 failed
```

Verified on a Book4 Ultra: mismatch refused, normal 1920x1080 YUYV setup unaffected, relay still reports a real picture. Existing suites unchanged — gst-tools 10/0, egl-vendor-pin 17/0, relay-loop-bench PASS.

Independent of #78; touches only `camera-relay-monitor.c` and adds one test file.

### DIFF
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -763,7 +763,12 @@ Run the installer to build it, or use 'camera-relay start' for always-on mode."
 
     # Clean up all children on exit
     cleanup_on_demand() {
-        pkill -P $$ 2>/dev/null
+        # `|| true` is load-bearing: with no children left, pkill exits 1, and a
+        # command failing inside an EXIT trap under set -e replaces the shell's
+        # exit status with its own. That silently turned every status into 1 —
+        # including the monitor's, which is the whole point of the sentinel
+        # above.
+        pkill -P $$ 2>/dev/null || true
         rm -f "$PID_FILE" "$STATE_CACHE"
     }
     trap cleanup_on_demand EXIT
@@ -795,8 +800,11 @@ Run the installer to build it, or use 'camera-relay start' for always-on mode."
                 info "Camera released, resuming idle"
                 ;;
         esac
-    done < <("$MONITOR_BIN" "$loopback_dev" 1920 1080 \
-             -- "${gst_cmd[@]}" 2>"$gst_log"; echo "__monitor_exit:$?")
+    # `|| rc=$?` rather than a bare `; echo ...$?`: the subshell inherits set -e,
+    # so a failing monitor would abort it before the sentinel was ever emitted —
+    # which looks exactly like the bug this is here to prevent.
+    done < <(rc=0; "$MONITOR_BIN" "$loopback_dev" 1920 1080 \
+             -- "${gst_cmd[@]}" 2>"$gst_log" || rc=$?; echo "__monitor_exit:$rc")
 
     if (( monitor_rc != 0 )); then
         warn "Monitor exited with status $monitor_rc — see $gst_log"
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -770,8 +770,18 @@ Run the installer to build it, or use 'camera-relay start' for always-on mode."
 
     # The monitor manages the pipeline subprocess itself.
     # We just read its events for status tracking.
+    #
+    # A process substitution gives no way to observe the child's exit code —
+    # it lands in neither $? nor PIPESTATUS. Without the sentinel below this
+    # function always returned 0, so a monitor that gave up looked like a
+    # clean shutdown and Restart=on-failure never fired: the same shape as the
+    # "Already running" exit-0 bug described above already_running_healthy().
+    local monitor_rc=0
     while IFS= read -r event; do
         case "$event" in
+            __monitor_exit:*)
+                monitor_rc="${event#__monitor_exit:}"
+                ;;
             READY)
                 info "Monitor ready, device visible to apps"
                 ;;
@@ -786,9 +796,13 @@ Run the installer to build it, or use 'camera-relay start' for always-on mode."
                 ;;
         esac
     done < <("$MONITOR_BIN" "$loopback_dev" 1920 1080 \
-             -- "${gst_cmd[@]}" 2>"$gst_log")
+             -- "${gst_cmd[@]}" 2>"$gst_log"; echo "__monitor_exit:$?")
 
+    if (( monitor_rc != 0 )); then
+        warn "Monitor exited with status $monitor_rc — see $gst_log"
+    fi
     info "On-demand relay stopped"
+    return "$monitor_rc"
 }
 
 cmd_stop() {
--- a/camera-relay/camera-relay-monitor.c
+++ b/camera-relay/camera-relay-monitor.c
@@ -191,7 +191,13 @@ static int open_writer(const char *device, int width, int height,
 	}
 
 	/* S_FMT is allowed to adjust what it was given and report the result in
-	 * the same struct, so a successful ioctl is not confirmation. */
+	 * the same struct, so a successful ioctl is not confirmation.
+	 *
+	 * sizeimage is deliberately not compared: for a packed format like YUYV
+	 * it is derived from the geometry, so width/height/pixelformat agreeing
+	 * already pins it. That stops holding the moment this writes anything
+	 * planar or compressed — add the sizeimage check along with the format
+	 * if that ever happens. */
 	if (fmt.fmt.pix.pixelformat != V4L2_PIX_FMT_YUYV ||
 	    fmt.fmt.pix.width != (__u32)width ||
 	    fmt.fmt.pix.height != (__u32)height) {
@@ -429,6 +435,10 @@ int main(int argc, char *argv[])
 		return 1;
 	}
 	pid_t our_pid = getpid();
+	/* Distinguishes "asked to stop" from "gave up". A clean exit tells the
+	 * supervisor everything is fine and Restart=on-failure will not fire, so
+	 * any path that abandons the device has to say so here. */
+	int exit_status = 0;
 
 	/* Open writer and set up device */
 	int fd = open_writer(device, width, height, frame_size, black_frame);
@@ -688,6 +698,7 @@ int main(int argc, char *argv[])
 							"[monitor] "
 							"Re-open "
 							"failed!\n");
+						exit_status = 1;
 						running = 0;
 						break;
 					}
@@ -767,5 +778,5 @@ int main(int argc, char *argv[])
 	free(black_frame);
 	if (fd >= 0)
 		close(fd);
-	return 0;
+	return exit_status;
 }
--- a/camera-relay/camera-relay-monitor.c
+++ b/camera-relay/camera-relay-monitor.c
@@ -139,6 +139,17 @@ static int count_other_openers(dev_t dev_id, pid_t our_pid,
 	return count;
 }
 
+/* Render a FOURCC for humans: 'YUYV', 'MJPG', ... */
+static const char *fourcc(char buf[5], __u32 pixelformat)
+{
+	buf[0] = (char)(pixelformat & 0xff);
+	buf[1] = (char)((pixelformat >> 8) & 0xff);
+	buf[2] = (char)((pixelformat >> 16) & 0xff);
+	buf[3] = (char)((pixelformat >> 24) & 0xff);
+	buf[4] = '\0';
+	return buf;
+}
+
 /* Open device for writing, set format, write initial black frame.
  * Returns fd on success, -1 on failure. */
 static int open_writer(const char *device, int width, int height,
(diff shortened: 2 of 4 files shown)


Andycodeman (changes requested):
The change itself is right, and I appreciate you being straight about the provenance — "this is not a fix for a confirmed field failure, it is a silent-failure path found while chasing one" is exactly the framing that lets me review it on its merits instead of guessing. Checking `S_FMT`'s *return struct* rather than just its return code is the correct instinct: V4L2 adjusting a format and reporting success is a classic trap, and writing fixed-size YUYV into whatever it settled on is garbage to every reader.

Both call sites do handle the new `-1`. The error text is genuinely good — it names the format the device is actually on, which is the thing you'd want at 1am.

**One blocking problem, and it's the recovery half of the argument.**

> Both call sites already handle a negative return, and the unit carries `Restart=on-failure` with `RestartSec=5` — so the relay retries and recovers on its own once whatever held the device lets go, rather than sitting there streaming into the void.

The first clause is true. The second isn't — the monitor's exit status never reaches systemd.

`cmd_start_ondemand` runs the monitor inside a process substitution:

```bash
done < <("$MONITOR_BIN" "$loopback_dev" 1920 1080 -- "${gst_cmd[@]}" 2>"$gst_log")

info "On-demand relay stopped"
}
```

Process substitution gives no way to observe the child's exit code — it isn't in `$?` and isn't in `PIPESTATUS`. Verified:

```
$ bash -c 'while IFS= read -r l; do :; done < <(echo hi; exit 42); echo "$? / ${PIPESTATUS[*]}"'
0 / 0
```

And the function's last statement is `info "On-demand relay stopped"`, so it returns 0 regardless. `camera-relay start --on-demand` therefore always exits 0, and `Restart=on-failure` never fires.

Separately, the re-open call site returns 0 on its own account — `running = 0; break;` falls through to the normal cleanup and `return 0`. That's the path that runs *after* a pipeline cycle, i.e. precisely the moment another application has had the device and may have left it on a foreign format. So the site most likely to hit your new check is also the one that exits cleanest.

Net effect of merging as-is: instead of streaming garbage silently, the relay **exits and stays dead until someone restarts it by hand.** No worse a picture, but it trades one silent failure for another, and the PR reads as though it self-heals.

Worth saying plainly: this is the same shape as one of the bugs v0.3.56 fixed —

> the on-demand systemd service started, saw "already running", printed it and **exited 0**. `Restart=on-failure` reads a clean exit as success, so systemd never retried.

There's even a comment about it at `camera-relay:231`. Easy one to walk back into.

## The fix

A sentinel line inside the substitution fits your existing event-stream protocol and avoids the subshell problem a pipe would introduce. Verified:

```bash
local rc=0
while IFS= read -r event; do
    case "$event" in
        __monitor_exit:*) rc="${event#__monitor_exit:}" ;;
        START) ... ;;
    esac
done < <("$MONITOR_BIN" ... ; echo "__monitor_exit:$?")

info "On-demand relay stopped"
return "$rc"
```

```
monitor exits 42 → function returns 42   ← reaches systemd, Restart fires
monitor exits 0  → function returns 0
```

And the re-open path should exit non-zero rather than `return 0`, so the same machinery covers it.

## On the test

Hygiene is good — it skipped correctly here with the relay running:

```
– skipped: no idle v4l2loopback device to test against
```

so it does keep its hands off a device a live relay is driving, as you said. I did **not** verify your "4 passed" figure: reaching it means stopping the relay on my daily driver, and I wasn't willing to do that mid-session. The build assertion passes (`✓ builds clean with -Wall -Wextra`).

One small thing while you're in there: the check compares `pixelformat`, `width` and `height` but not `sizeimage`. For YUYV those are determined by the geometry, so it's consistent today — worth a word in the comment about why it's sufficient, so nobody later "fixes" it by adding a format where it isn't.

Sort the exit-status propagation and I'll merge this.

4nrry:
You're right, and thanks for catching it — the recovery claim was the part I asserted instead of testing. Reproduced your one-liner before changing anything:

```
$ bash -c 'while IFS= read -r l; do :; done < <(echo hi; exit 42); echo "$? / ${PIPESTATUS[*]}"'
0 / 0
```

Fixed in a15ce4d.

**Exit status.** Sentinel inside the substitution, as you suggested — it fits the existing event protocol and dodges the subshell a pipe would bring:

```bash
done < <("$MONITOR_BIN" ... ; echo "__monitor_exit:$?")
```

`cmd_start_on_demand` now ends with `return "$monitor_rc"`, and `cmd_start` already forwards it with a bare `return`.

**The re-open path.** You were right that it was the worse of the two, and for the reason you gave: it runs right after a pipeline cycle, so it is the site most likely to trip the new check and it was the one exiting cleanest. `main()` now carries an `exit_status` that path sets, instead of falling through to `return 0`.

**sizeimage.** Comment added. Short version: for a packed format like YUYV it follows from the geometry, so width/height/pixelformat agreeing already pins it — and it says explicitly that this stops holding for anything planar or compressed, so the next person doesn't have to work it out.

**Test.** `tests/test-monitor-exit-propagation.sh`. It lifts the read loop out of `camera-relay` with `sed` rather than retyping it, so a stale copy can't pass. Checked that it actually fails: deleting the sentinel gives

```
✗ monitor exit status was swallowed — the process substitution hides it
3 passed, 1 failed
```

and restoring it goes back to 4/4.

On your note about not verifying the "4 passed" figure — that's fair, and I'd rather you didn't stop your relay for it. The skip path is the one that matters for safety anyway, and you confirmed that one yourself.

Suites here: exit-propagation 4/4, writer-format 4/4 (1/1 with a live relay holding the loopback), gst-tools 10/0, egl-vendor-pin 17/0, relay-loop-bench PASS.

One thing worth flagging: this now touches `camera-relay` as well as the monitor, and #78 modifies the same function. They do **not** conflict — `git merge-tree` reports zero, since #78 changes how `gst_cmd` is built and this changes the read loop below it — but they're no longer strictly independent, so merge order is worth a glance.

4nrry:
Follow-up: I went to install the monitor on the machine and test it for real, and the status still never arrived. Two more things were eating it — both of which defeated the sentinel completely, so a15ce4d alone did not actually fix what you flagged. Fixed in 14bb7a1.

**`set -e` inside the process substitution.** The subshell inherits it, so a failing monitor aborted it *before* `echo "__monitor_exit:$?"` ever ran. The loop saw only `READY`:

```
DEBUG evento=[READY]
DEBUG monitor_rc=[0]
```

Capturing through an `||` branch keeps `set -e` out of the way:

```bash
done < <(rc=0; "$MONITOR_BIN" ... || rc=$?; echo "__monitor_exit:$rc")
```

Worth noting the snippet in your review has the same issue — it works in a plain shell, but not under this script's `set -euo pipefail`. Easy one to miss; I copied it straight across and it took a live run to surface.

**The EXIT trap, which gets the last word.** Under `set -e`, a command failing inside a trap replaces the status the shell was exiting with. `cleanup_on_demand`'s `pkill -P $$` exits 1 whenever no children are left — the normal case:

```
$ bash -c 'set -e; cleanup() { pkill -P $$ 2>/dev/null; rm -f /tmp/x; }
           trap cleanup EXIT; f() { return 42; }; f'; echo $?
1
```

So every status became 1 regardless of the sentinel. That one is pre-existing — it has been flattening exit codes since before this PR — and `|| true` on the pkill fixes it.

**Why my test didn't catch either.** Its subshell used `set -uo pipefail` while `camera-relay` uses `set -euo pipefail`. Without `-e` the sentinel is emitted either way, so the test happily agreed with a script that was broken in production. It now matches the script's flags and covers the trap too. (The trap case runs as its own shell — `$$` is unchanged in a subshell, so `pkill -P $$` would otherwise kill the test harness. Learned that one the noisy way.)

End-to-end now, against a merged #78+#79 relay driven by a stub monitor:

```
monitor exits 42 -> camera-relay start --on-demand exits 42
monitor exits 0  -> exits 0
```

Suites: exit-propagation 5/5, writer-format 4/4 (1/1 with a live relay holding the loopback), gst-tools 10/0, egl-vendor-pin 17/0, shellcheck clean.

The lesson I'm taking from this one: I claimed recovery worked, you caught that I hadn't tested it, I fixed the mechanism — and it still didn't work, for two reasons neither of us saw by reading. It only fell out of running it against a real relay.

Andycodeman (approved):
Merging. And to put the important part first: **the fix I gave you was wrong, and you caught it by running it.** `set -e` inside the process substitution aborts the subshell before the sentinel `echo` ever executes, so the snippet in my review would have looked correct, passed a casual test, and shipped the same swallowed status. That's on me, and finding it required exactly the step I hadn't taken.

Worth recording how nearly I missed it a second time. Checking your claim, my first attempt used `false` — a builtin — and the sentinel appeared, which had me doubting the diagnosis. It only reproduces with an **external** command, which is the real case:

```
plain sentinel, external failing command → saw: READY            (no sentinel at all)
rc=0; cmd || rc=$?; echo                 → saw: __monitor_exit:42
```

So your `|| rc=$?` form isn't just belt-and-braces, it's load-bearing. Good comment on it too.

**The EXIT trap one is the better find of the two**, and it wasn't yours to fix. `pkill -P $$` exits 1 with no children left, and under `set -e` a command failing inside a trap replaces the status the shell was exiting with. Confirmed:

```
without || true → exit 1      with || true → exit 42
```

That has been flattening every exit code in this script since long before your PR, and it would have silently defeated the sentinel no matter how correctly it was written. Nice catch, and the comment explaining why the `|| true` is load-bearing is what stops it being "cleaned up" later.

**On the tests.** I mutation-tested rather than taking the numbers on trust — both guards are real:

| mutation | result |
|---|---|
| revert to the naive sentinel (my version) | `✗ monitor exit status was swallowed` — 4/1, exit 1 |
| drop `\|\| true` from the pkill | `✗ the EXIT trap rewrote the exit status to 1` — 4/1, exit 1 |

Lifting the read loop out of `camera-relay` with `sed` rather than retyping it is the right call — a hand-copied loop would have drifted and passed forever. And matching the script's exact `set -euo pipefail` after the first version used `set -uo pipefail` is precisely the bug your test had; good that it now can't recur.

`test-writer-format-check.sh` skipped correctly here with a live relay holding the loopback, so the safety path is confirmed on my side too. I still haven't exercised your 4/4 figure and I'm content not to.

Also noted and agreed on merge order: this and #78 touch the same function without conflicting. #79 goes in first.

The summary you wrote is the part I'd keep:

> I claimed recovery worked, you caught that I hadn't tested it, I fixed the mechanism — and it still didn't work, for two reasons neither of us saw by reading. It only fell out of running it against a real relay.

That's the whole lesson, and it applies to my review as much as your patch. Thanks for chasing it to the end rather than stopping at the first green run.

PR #80, opened 2026-08-04
webcam-fix-libcamera: two bugs the uninstall/reinstall cycle exposed

4nrry:
You asked in #78 for `uninstall.sh` to be exercised rather than just reordered. It was, end to end, followed by a clean `install.sh`. Both scripts completed successfully and both left something broken.

## 1. The uninstaller never restored node ownership

The reorder in #78 was correct about *when* to revert, but relied on a udev behaviour that does not exist. Removing the rule and triggering `change` re-runs the rules — properties and the `uaccess` tag come back — but **udev does not re-apply node ownership on a change event**. `50-udev-default.rules` sets `GROUP="video"` unconditionally for `video4linux`, and the nodes still kept the `camera-relay` GID. `udevadm settle` makes no difference; it is not a reload race.

The uninstaller then deleted the group, so all 48 nodes ended up owned by a GID that no longer resolves:

```
/dev/video1  -> root:UNKNOWN (gid 974)
nodes with unresolvable gid: 48
```

Worse than cosmetic: if 974 is ever reused, those nodes silently become reachable by whatever group gets it. And nothing looks wrong at the time — the `uaccess` ACL returns once the rule is gone, so the camera keeps working.

Ownership is now restored explicitly instead of delegated to udev, matched by GID rather than by name so the same code repairs a system already left dangling. `groupdel` is guarded on no node still referencing the group. Verified on the machine that was in exactly that state:

```
[0/8] Restoring raw camera node ownership...
  ✓ Raw nodes back to the 'video' group (48 restored)

/dev/video1 -> root:video (gid 44)
nodes with unresolvable gid: 0
```

## 2. Reinstalling drops /usr/local from the linker path

`/etc/ld.so.conf.d/libcamera-local.conf` is written inside the source-build branch, which is skipped whenever an adequate libcamera is already installed. `uninstall.sh` removes it unconditionally. So uninstall → install leaves it gone.

The failure is delayed, which is what makes it worth fixing rather than shrugging at. The ldconfig cache still holds the old entries, so `ldconfig -p` looks right and everything works. It breaks at the next unrelated `ldconfig` — a package upgrade, another installer — after which the **system** libcamera wins. `camera-relay-gst` deliberately does not propagate `LD_LIBRARY_PATH`, precisely because it is setgid, so it would load the unpatched build with nothing pointing at the cause.

Now written whenever a libcamera exists under `/usr/local`, regardless of who installed it.

## Cycle result

Full state diff, before the uninstall versus after the reinstall, is now identical except for `v4l2loopback` not being loaded in the session — the module config is in place and it loads at boot.

Neither bug reproduces from either script alone, and neither is visible in review: in both cases the broken state looks healthy from the outside. #78 closed the ordering hole; running the thing is what found what was behind it.

## Unrelated, but surfaced by the same run

Both scripts restart PipeWire (`install.sh` step 14, `uninstall.sh` step 8). Any app playing audio loses its stream, which presents to the user as "the camera fix broke my speakers" — that is the conclusion I reached here before checking. Worth its own issue; I have not touched it.

### DIFF
--- a/webcam-fix-libcamera/install.sh
+++ b/webcam-fix-libcamera/install.sh
@@ -1131,6 +1131,30 @@ if ! verify_libcamerasrc; then
     exit 1
 fi
 
+# Whoever put a libcamera under /usr/local, the linker has to prefer it — the
+# source build above writes this file, but that branch is skipped when a good
+# enough libcamera is already installed. uninstall.sh removes the file
+# unconditionally, so an uninstall/reinstall cycle used to drop /usr/local out
+# of the search path and leave the *system* libcamera winning.
+#
+# Nothing breaks at that moment, because the ldconfig cache still holds the old
+# entries. It breaks at the next unrelated `ldconfig` — a package upgrade,
+# another installer — and then camera-relay-gst, which deliberately does not
+# propagate LD_LIBRARY_PATH, silently loads the unpatched build.
+if compgen -G "/usr/local/lib*/libcamera.so.*" > /dev/null \
+   || compgen -G "/usr/local/lib/*/libcamera.so.*" > /dev/null; then
+    if [[ ! -f /etc/ld.so.conf.d/libcamera-local.conf ]]; then
+        sudo tee /etc/ld.so.conf.d/libcamera-local.conf > /dev/null << 'EOF'
+/usr/local/lib
+/usr/local/lib64
+/usr/local/lib/x86_64-linux-gnu
+/usr/local/lib/aarch64-linux-gnu
+EOF
+        sudo ldconfig
+        echo "  ✓ Restored /usr/local to the dynamic linker search path"
+    fi
+fi
+
 # ──────────────────────────────────────────────
 # [10/14] Install PipeWire libcamera plugin
 # ──────────────────────────────────────────────
--- a/webcam-fix-libcamera/uninstall.sh
+++ b/webcam-fix-libcamera/uninstall.sh
@@ -32,7 +32,34 @@ sudo rm -f /etc/udev/rules.d/72-camera-relay-mc-nodes.rules \
            /etc/udev/rules.d/90-camera-relay-mc-nodes.rules
 sudo udevadm control --reload-rules 2>/dev/null || true
 sudo udevadm trigger --action=change --subsystem-match=video4linux 2>/dev/null || true
-echo "  ✓ Raw nodes back to their default ownership"
+sudo udevadm settle 2>/dev/null || true
+
+# The trigger above is not enough on its own. It re-runs the rules — the
+# properties and the uaccess tag come back — but udev does not re-apply node
+# ownership on a change event, so every node keeps the camera-relay GID. With
+# the group deleted further down that leaves a dangling numeric GID, and if
+# that number is ever reused the nodes silently become reachable again.
+#
+# So restore the group explicitly rather than hoping udev does it. Nodes are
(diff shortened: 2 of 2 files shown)


Andycodeman (approved):
Merging. You went and ran the cycle I explicitly said not to push yourself on, and it turned up two bugs neither of us would have found by reading. That was the right call and I'm glad you made it.

**Bug 1 — your udev diagnosis is exactly right, and it took me three attempts to confirm it.** Worth writing out, because the behaviour is subtle enough that a reviewer could easily "disprove" it and wave the fix away.

My first test looked like it contradicted you:

```
manual chgrp, then --action=change   → NOT restored
```

but that measures something else — udev doesn't correct out-of-band changes it didn't make. So I tried again with a rule actually setting `GROUP=`, and it looked like it contradicted you the *other* way:

```
rule present + --action=change       → rule applied
```

At which point I was about to report `install.sh:1623` as buggy for using `change` to reassign the group. **It isn't, and your comment on that line is accurate** — I'd have filed a false finding if I'd stopped there.

Your actual scenario is the one that behaves differently — rule *removed*, then triggered:

```
rule present + add       → audio   (rule applied)
rule REMOVED + change    → audio   ← ownership NOT restored
rule REMOVED + add       → video   (restored)
```

So udev applies ownership a rule *specifies*, but won't revert to the default on a `change` event once the rule is gone. Exactly as you described, and `udevadm settle` is indeed irrelevant — it isn't a race.

For the record, `--action=add` would also have restored it, and would have been a one-line fix. I think your explicit `chgrp` is the better choice anyway: matching by GID rather than by name means the same code repairs a system already left dangling by an earlier run, which a trigger can't do once the group is gone. Guarding `groupdel` on nothing still referencing the group is the right belt-and-braces.

The "silently reachable if the GID is reused" point is the part that makes this more than tidiness, and I agree it's the right severity read.

**Bug 2 — confirmed structurally.** `install.sh:825` writes `libcamera-local.conf` inside the source-build branch (the comment directly above it says as much), and `uninstall.sh:181` removes it unconditionally. Your delayed-failure analysis holds: the ldconfig cache masks it until some unrelated `ldconfig`, after which `camera-relay-gst` loads the system build with nothing pointing at the cause — precisely because it doesn't propagate `LD_LIBRARY_PATH`, which is a property we deliberately gave it in #78.

My own machine is one of the affected shape: `/etc/ld.so.conf.d/libcamera-local.conf` present with a `/usr/local` libcamera behind it. So this would have bitten a real system, not a hypothetical one.

`bash -n` clean on both scripts.

## One question, not blocking

The restore block sits at ~1150, after `verify_libcamerasrc` exits on failure at 1118. A machine that has *already* hit the delayed failure — unrelated `ldconfig` ran, system libcamera now winning — is exactly the machine whose owner will re-run the installer to fix it, and the verify may bail before the restore ever executes. Moving the block above the verify would let re-running the installer be the cure rather than a dead end.

I don't know for certain that the verify fails in that state (`libcamerasrc` may well still load, just against the unpatched build), which is why it's a question rather than a finding. If you know it passes, leave it.

## The PipeWire restart

Agreed it wants its own issue, and thanks for not folding it in. Your framing is the useful bit — *"the camera fix broke my speakers"* is what the user reports, and the honest note that you reached that conclusion yourself before checking is what makes it worth filing rather than dismissing.

Cutting a release with this in it. Both of these were silent — a dangling GID and a linker path that breaks later — so they shouldn't sit in a shipped tag.

Issue #65, opened 2026-07-09
Webcam dosen't work on GalaxyBook5 Pro 360 960QHA


seshf:
Kernel: 7.1.1-070101-generic (64-bit)
Wayland
Kubuntu 26.04 LTS

xxx@xxx-960qha:~$  cam -l
[0:05:05.767199690] [10627]  INFO Camera camera_manager.cpp:340 libcamera v0.7.0+dirty (2026-07-09T02:23:10CEST)
[0:05:05.802548745] [10628] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 3-0010': Unable to get rectangle 2 on pad 0/0: Inappropriate ioctl for device
[0:05:05.802655716] [10628]  WARN CameraSensor camera_sensor_legacy.cpp:402 'ov02e10 3-0010': The PixelArraySize property has been defaulted to 1928x1088
[0:05:05.802683653] [10628] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 3-0010': Unable to get rectangle 1 on pad 0/0: Inappropriate ioctl for device
[0:05:05.802703895] [10628]  WARN CameraSensor camera_sensor_legacy.cpp:413 'ov02e10 3-0010': The PixelArrayActiveAreas property has been defaulted to (0, 0)/1928x1088
[0:05:05.802734982] [10628] ERROR V4L2 v4l2_subdevice.cpp:1192 'ov02e10 3-0010': Unable to get rectangle 0 on pad 0/0: Inappropriate ioctl for device
[0:05:05.802753280] [10628]  WARN CameraSensor camera_sensor_legacy.cpp:421 'ov02e10 3-0010': Failed to retrieve the sensor crop rectangle
[0:05:05.802769616] [10628]  WARN CameraSensor camera_sensor_legacy.cpp:427 'ov02e10 3-0010': The sensor kernel driver needs to be fixed
[0:05:05.802784205] [10628]  WARN CameraSensor camera_sensor_legacy.cpp:429 'ov02e10 3-0010': See Documentation/sensor_driver_requirements.rst in the libcamera sources for more information
[0:05:05.804795005] [10628]  WARN CameraSensorProperties camera_sensor_properties.cpp:538 No static properties available for 'ov02e10'
[0:05:05.804829518] [10628]  WARN CameraSensorProperties camera_sensor_properties.cpp:540 Please consider updating the camera sensor properties database
[...]

Issue #54, opened 2026-05-16
Camera OV02C10 not working — Galaxy Book4 Pro (940XGK)


Bernardo-Lebron:
After running the updated webcam-fix script and rebooting, the camera still doesn't work.

**dmesg:**
ov02c10: unknown parameter 'clk_freq' ignored
ov02c10 i2c-OVTI02C1:00: error -EINVAL: external clock 26000000 is not supported
ov02c10 i2c-OVTI02C1:00: probe with driver ov02c10 failed with error -22

**System info:**
- Kernel: 6.17.0-23-generic
- Distro: Zorin OS 18.1
- Subsystem ID: 0x144dca07

**Modules loaded:** intel_ipu6, intel_ipu6_isys, ov02c10, ivsc_csi, ivsc_ace, mei_vsc

The BIOS/ACPI reports 26MHz but the ov02c10 driver only accepts 19.2MHz or 24MHz. The `clk_freq` module parameter is not recognized by this driver version.

Andycodeman:
Thanks for the detailed report — the logs make this clear.

This is the **26 MHz sensor-clock issue**. Your 940XGK's ACPI provides a 26 MHz clock to the OV02C10, but the in-tree `ov02c10` driver only accepts 19.2 MHz, so probe fails with `-EINVAL` (`-22`). The repo has a DKMS-patched driver for exactly this: [`ov02c10-26mhz-fix/`](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/tree/main/ov02c10-26mhz-fix).

Two things in your dmesg:

1. **`ov02c10: unknown parameter 'clk_freq' ignored`** — this is a red herring, *not* the cause. There is no `clk_freq` module parameter in this (or the upstream) driver. It comes from a leftover `/etc/modprobe.d/*.conf` from some older workaround. Harmless, but worth removing.

2. **`external clock 26000000 is not supported`** *after* you ran the fix and rebooted means the **stock in-tree driver is still the one binding** — the patched DKMS module isn't winning. On Zorin/Ubuntu the usual reason is **Secure Boot**: an unsigned DKMS module is silently rejected, and the kernel falls back to the distro-signed in-tree driver (which still rejects 26 MHz). A silent DKMS build failure on the fresh 6.17 kernel can do the same.

[...]

PR #83, opened 2026-08-04
camera-relay: make Chromium-family browsers actually see the camera

4nrry:
Chrome, Chromium, Brave, Edge and every Electron app list **no camera at all** after install, while Firefox works on the first try. Two independent bugs were stacked behind that, and **each one alone fully explains the symptom** — which is how #54 and #65 both landed on the wrong mechanism.

Reproduced on a Galaxy Book4 Ultra, Ubuntu 26.04, kernel 7.0.0-28, libcamera 0.7.2, v4l2loopback 0.15.3, Google Chrome 151 (deb — not snap, so sandboxing is not a factor), relay running and healthy.

```js
// before
await navigator.mediaDevices.enumerateDevices()
// { videoinput: [], audioinput: 1, audiooutput: 1 }   ← audio enumerates fine
```

Not a permission problem (state was `prompt`), not the sandbox, not the relay.

## Bug 1 — Chromium filters the relay node out of V4L2 enumeration

[`video_capture_device_factory_v4l2.cc`](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/media/capture/video/linux/video_capture_device_factory_v4l2.cc) accepts a node only when it reports capture and **not** output:

```c
(cap.capabilities & V4L2_CAP_VIDEO_CAPTURE &&
 !(cap.capabilities & V4L2_CAP_VIDEO_OUTPUT)) ||
(cap.capabilities & V4L2_CAP_DEVICE_CAPS &&
 cap.device_caps & V4L2_CAP_VIDEO_CAPTURE &&
 !(cap.device_caps & V4L2_CAP_VIDEO_OUTPUT))
```

Under `exclusive_caps=0` the relay reports both (`caps 0x85200003`, `device caps 0x05200003`), so both branches fail and the node is dropped before any prompt. Firefox accepts dual caps and reads the node directly — the whole reason the two disagree.

**Corrections to the record:**

- **#54** attributed this to the udev `ID_V4L_CAPABILITIES` property. Chromium enumerates with `base::FileEnumerator` over `/dev/video*` and never reads udev — the property was already `:capture:` on the reproducing machine and Chrome still saw nothing. `70-camera-relay-capabilities.rules` is **kept** (other udev consumers do read it) but its comments no longer claim it does anything for Chromium.
- **#65** called dual caps *"a strong inference, not something confirmed from Chromium's side"*. Now confirmed from the source. It also understated the scope: the check is unconditional, so this is not a "some setups" problem and the flag is **required**, not a fallback.

**Fix:** `camera-relay/chromium-pipewire-camera.sh` enables `chrome://flags/#enable-webrtc-pipewire-camera` across Chrome, Chromium (incl. snap), Brave, Vivaldi and flatpaks. Gated on libcamera 0.7+, refuses while the browser is running (Chromium rewrites `Local State` on exit and would silently discard the write), backs the file up, byte-idempotent. Both installers call it; both uninstallers call `disable`.

> The top-level README already claimed *"the installer automatically enables the PipeWire camera flag"*. No installer did — the only matches were `echo` lines telling the user to do it by hand. Now one does.

`exclusive_caps=1` would fix this at the source with no flag, but that reopens the WirePlumber classification problem this project deliberately moved away from. Documented as a trade-off rather than changed.

## Bug 2 — WirePlumber caches a stale format list, so the flag alone changes nothing

Enabling the flag did **not** make the camera appear. The flag was working:

```
camera_portal.cc:215]    Camera access granted by the XDG portal.
pipewire_session.cc:99]  Found Camera: Camera Relay (V4L2)
```

Chrome *finds* the camera and offers no device, because the node's formats are stale:

```
pw-cli enum-params <node> EnumFormat     v4l2-ctl -d /dev/video0 --list-formats-ext
  VideoFormat:BGRx                         [0]: 'YUYV' (YUYV 4:2:2)
  Rectangle 2x1 → 8192x8192 (Range)             Size: Discrete 1920x1080
```

`v4l2loopback` loads at boot with no producer and advertises its catch-all range. The monitor pins `YUYV 1920x1080` only at login, and `camera-relay.service` is ordered `After=wireplumber.service` — so WirePlumber always probes the unconfigured device and never looks again. WebRTC needs a **discrete** rectangle out of `SPA_PARAM_EnumFormat`; a range yields no resolution, so the camera is built with zero capabilities. Firefox is immune: it never consults PipeWire's format list.

Confirmed by an actual reboot mid-investigation — the flag survived, the node regressed to `BGRx / Rectangle 2x1`, Chrome went back to finding nothing.

**Fix:** `nudge_wireplumber` returns, from `ExecStartPost`. The two reasons it was removed do not apply where it now runs:

- *"sudo prompt blocks/confuses users"* — that was `udevadm trigger`, needed only because the legacy `v4l2-relayd` is a **system** service. `camera-relay` is a user service, so `systemctl --user` needs no privileges, and the restart alone is sufficient.
- *"restart disrupts active camera streams"* — nothing is streaming at relay start.

It waits for the monitor to pin the format (`ExecStartPost` runs as soon as a `Type=simple` `ExecStart` *forks*, which is too early — the first attempt read no format and silently did nothing), and restarts WirePlumber **only** when the formats disagree, so a manual relay restart costs no audio glitch. The `-` prefix keeps a browser-only correction from ever taking the relay down.

This matches the conclusion already recorded in `webcam-fix/README.md` for the legacy stack: *"`device.capabilities` is read-only in PipeWire … The only fix is to make WirePlumber re-discover the device after the relay attaches."* There is no lighter re-probe — the V4L2 monitor is udev-driven and WirePlumber exposes no per-device reprobe call.

## Result

```json
{ "videoinput": ["Camera Relay (V4L2)"],
  "frame": "1920x1080", "frameRate": 30,
  "distinctColors": 2004, "verdict": "REAL PICTURE" }
```

## Also

- `camera-relay doctor` now **evaluates** Chromium's capability expression against the live node, compares PipeWire's advertised format against the device's, and reports the flag state per browser profile — instead of suggesting the user try things.
- `Chrome | Working | Works via V4L2 camera relay` in the compatibility tables was the opposite of the truth; corrected everywhere. Edge is now listed as not working, since no flag exists to work around it there.
- Findings recorded in `docs/triage/issue-54-findings.md` and `issue-65-findings.md`.

## Verification

- Boot race simulated end to end: stop relay → restart WirePlumber (node goes to `BGRx 2x1`) → start relay → `YUY2 1920x1080`, relay still active.
- 63 new assertions across two suites (`test-chromium-pipewire-flag.sh`, `test-wireplumber-format-nudge.sh`); **125 total, all passing** across the seven suites in `camera-relay/tests/`.
- `bash -n` clean on every touched script. `shellcheck -S warning` reports **no new warnings versus `main`**: the three new scripts are at zero, and each pre-existing one carries exactly the count it already had on `main` (`camera-relay` 1, `webcam-fix-book5/install.sh` 7, `webcam-fix-libcamera/install.sh` 9, `webcam-fix-book5/uninstall.sh` 1).
- Rebased onto `main` (`4dc3825`). The branch previously sat on `1f30ca6`, before the setgid pipeline launcher landed, so its `camera-relay` predated that work — running this branch's `install.sh` would have reinstalled a relay that invokes `gst-launch-1.0` directly instead of `camera-relay-gst`, leaving it unable to open the raw nodes. The rebase resolves cleanly and touches none of that code; the numbers above are measured post-rebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,13 +59,13 @@ curl -sL https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/archive/
 
 To uninstall: `./uninstall.sh && sudo reboot`
 
-The webcam works with **Firefox, Chromium, Zoom, Teams, OBS, mpv, VLC**, and most other apps. For Chromium-based browsers (Brave, Chrome), the installer automatically enables the PipeWire camera flag.
+The webcam works with **Firefox, Chromium, Zoom, Teams, OBS, mpv, VLC**, and most other apps. Chromium-based browsers (Chrome, Brave) cannot see the camera relay over V4L2 and need the PipeWire camera flag; the installer sets it for you when the browser is closed during install and libcamera is 0.7+.
 
 ### Webcam Fix (built-in camera not detected) — Lunar Lake / Galaxy Book5 / Arch, Fedora & Ubuntu
 
 > Confirmed working on Samsung Galaxy Book5 Pro 940XHA (Fedora 43), 960XHA (Ubuntu 24.04), Dell XPS 13 9350 (Arch), and Lenovo X1 Carbon Gen13 (Fedora). See the [full README](webcam-fix-book5/) for details, known issues, and tested hardware.
 
-> **Requires kernel 6.18+** and **Arch, Fedora, or Ubuntu** (Ubuntu requires libcamera 0.5.2+ and kernel 6.18+ built from source). Includes an on-demand camera relay for non-PipeWire apps (Zoom, OBS, VLC) with near-zero idle CPU usage. For Chromium-based browsers (Brave, Chrome), the installer automatically enables the PipeWire camera flag.
+> **Requires kernel 6.18+** and **Arch, Fedora, or Ubuntu** (Ubuntu requires libcamera 0.5.2+ and kernel 6.18+ built from source). Includes an on-demand camera relay for non-PipeWire apps (Zoom, OBS, VLC) with near-zero idle CPU usage. Chromium-based browsers (Chrome, Brave) cannot see that relay over V4L2 and need the PipeWire camera flag; the installer sets it for you when the browser is closed during install and libcamera is 0.7+.
 
 > **OV02E10 purple tint fix:** Samsung Book5 models with the OV02E10 sensor mounted upside-down get purple/magenta tint due to a bayer pattern mismatch after the rotation flip. A patched libcamera build fixes this — see [OV02E10 bayer fix](webcam-fix-book5/libcamera-bayer-fix/) and the [webcam-fix-book5 README](webcam-fix-book5/) for details.
 
@@ -130,7 +130,7 @@ The built-in webcam uses Intel IPU6 (Meteor Lake or Raptor Lake) with an OmniVis
 
 - PipeWire-native apps (Firefox, Chromium) access the camera directly — no relay needed
 - Non-PipeWire apps use the on-demand V4L2 relay: near-zero CPU when idle, camera activates only when an app opens the device
-- Chromium browser PipeWire camera flags are auto-enabled during install
+- Chromium-family browsers are filtered out of the relay's V4L2 node by their own capability check, so the installer enables the PipeWire camera flag for them (browser must be closed; libcamera 0.7+)
 
 > **Multi-distro:** Supports **Ubuntu, Fedora, and Arch-based distros**. The install script auto-detects your distro. Galaxy Book5 (Lunar Lake / IPU7) is not supported (different driver stack) — see [webcam-fix-book5](webcam-fix-book5/).
 
@@ -140,7 +140,7 @@ For Galaxy Book5 (Lunar Lake / IPU7) on Arch, Fedora, and Ubuntu (source build).
 
 - PipeWire-native apps (Firefox, Chromium) access the camera directly
 - Non-PipeWire apps (Zoom, OBS, VLC) use the on-demand V4L2 relay: near-zero CPU when idle, camera activates only when an app opens the device
-- Chromium browser PipeWire camera flags are auto-enabled during install
+- Chromium-family browsers are filtered out of the relay's V4L2 node by their own capability check, so the installer enables the PipeWire camera flag for them (browser must be closed; libcamera 0.7+)
 
 For Samsung Book5 models with the OV02E10 sensor, an additional [patched libcamera build](webcam-fix-book5/libcamera-bayer-fix/) is needed to fix the purple/magenta tint caused by bayer pattern mismatch after rotation flip. See the [webcam-fix-book5 README](webcam-fix-book5/) for details.
 
--- a/camera-relay/camera-relay
+++ b/camera-relay/camera-relay
@@ -634,13 +634,149 @@ stop_pipeline() {
     fi
 }
 
-# No longer needed: with exclusive_caps=0, v4l2loopback always has
-# VIDEO_CAPTURE, so WirePlumber discovers it correctly on its own.
-# The old nudge_wireplumber (udevadm trigger + WirePlumber restart) was
-# needed for exclusive_caps=1 where caps only flipped after the relay
-# opened the device for writing. Removed because:
-# 1. sudo prompt blocks/confuses users
-# 2. WirePlumber restart disrupts active camera streams (causes flip)
+# nudge_wireplumber — make WirePlumber re-read the loopback's formats.
+#
+# This was removed once, on the grounds that exclusive_caps=0 always advertises
+# VIDEO_CAPTURE so WirePlumber classifies the node correctly on its own. That
+# part is true. What it missed is that *capabilities* are not the only thing
+# WirePlumber reads once and caches — the format list is too.
+#
+# v4l2loopback is loaded at boot from modules-load.d with no producer, so it
+# advertises its generic catch-all format set: BGRx/xRGB/BGR at any size from
+# 2x1 to 8192x8192, expressed as a Choice:Range. The relay's monitor only pins
+# YUYV 1920x1080 when it starts, at login — and camera-relay.service is
+# ordered After=wireplumber.service, so WirePlumber has already probed the
+# unconfigured device and will never look again.
+#
+# Firefox never noticed: it reads /dev/video0 directly. WebRTC's PipeWire camera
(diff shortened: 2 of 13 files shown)


Andycodeman (approved):
Excellent work. Both diagnoses are correct, and I checked each one independently rather than taking the writeup's word for it. Merging this.

## What I verified

**Bug 1 — confirmed from Chromium's side, as you claim.** I pulled current `video_capture_device_factory_v4l2.cc`. `GetDevicesInfo()` enumerates with `base::FileEnumerator(path, false, FILES, "video*")` — no udev in that path at all — and rejects any node reporting `VIDEO_CAPTURE` and `VIDEO_OUTPUT` together as memory-to-memory. My relay reports exactly the numbers you quote:

```
Capabilities     : 0x85200003     Video Capture + Video Output + ...
Device Caps      : 0x05200003     Video Capture + Video Output + ...
```

So the correction to #54 stands, and keeping `70-camera-relay-capabilities.rules` while rewriting its comments is the right call.

**Bug 2 — reproduced on my Book4 Ultra.** Not hypothetical: my machine is in the broken state right now, relay healthy and up for five days.

```
$ v4l2-ctl -d /dev/video0 --list-formats-ext
	[0]: 'YUYV' (YUYV 4:2:2)
		Size: Discrete 1920x1080

$ pw-cli enum-params 38 EnumFormat
      Id 8        (Spa:Enum:VideoFormat:BGRx)
      Choice: type Spa:Enum:Choice:Range, flags 00000000 40 8
        Rectangle 2x1
        Rectangle 2x1
        Rectangle 8192x8192
```

Also worth recording, since your fix depends on it: the device reports `1920x1080` **discrete while the relay is idle** — pipeline stopped, no client. So the monitor really does pin the format at startup and the bounded wait converges instead of always timing out.

**The `ExecStartPost` path, end to end on real hardware.** I extracted `loopback_real_format` / `pipewire_node_for` / `nudge_wireplumber` / `cmd_nudge_wireplumber` from the branch and ran them against the live device with only `systemctl` stubbed:

```
  [info] WirePlumber has stale formats for /dev/video0 (sees 2x1, device offers 1920x1080)
  [info] Restarting WirePlumber so PipeWire-based apps (Chrome, Brave) see a usable camera
rc=0
--- stubbed systemctl calls ---
STUB systemctl --user restart --no-block wireplumber
```

Correct detection, correct invocation, exit 0 — and no client-connect event in the relay journal, so the polling does not wake the camera at every login. (`doctor` does wake it, but that is pre-existing: it streams 60 frames off the loopback.)

**Tests:** 7 suites, 126 assertions, 0 failures here (41 + 18 + 10 + 29 + 5 + 22 + 1; two skips in `test-gst-tools-check.sh` because a real relay holds the loopback). `bash -n` clean on all 8 touched scripts. Rebase verified — `pr83^` *is* `4dc3825`.

**And the README really was lying.** On `main` the only `enable-webrtc-pipewire-camera` matches in either installer are `echo` lines. Fair jab.

---

## One thing I'm fixing in a follow-up rather than blocking on

`chromium-pipewire-camera.sh` is executed from `$RELAY_DIR` but never installed. Both installers `cp` `camera-relay`, the monitor, the systray, the `.desktop` and the icons into system paths — the new script stays in the extracted tarball.

That bites harder than it looks, because the running-browser guard makes "skipped, re-run it later" the *common* case, and all three places that tell the user how to re-run it point at a relative path:

- the installers' closing message: `quit it and run: camera-relay/chromium-pipewire-camera.sh`
- both READMEs: `cd camera-relay && ./chromium-pipewire-camera.sh`
- `doctor`: `Set it with: camera-relay/chromium-pipewire-camera.sh`

...into a tree the documented `curl -sL …/archive/main.tar.gz | tar xz` flow invites people to delete. I'll land the `install -m 755` plus the reference updates myself so this ships now.

## Non-blocking

1. **`enable` silently reverses an explicit "Disabled".** My own Brave held `enable-webrtc-pipewire-camera@2` — set deliberately, quite possibly following this project's own earlier "leave it OFF" advice. It gets rewritten to `@1` and reported as a plain `✓ Brave: PipeWire camera flag enabled`. On 0.7+ that's the right end state and the backup is taken first, but it should say a preference was reversed. `enable_case "explicitly disabled"` currently locks in the silent behaviour.

2. **Two vocabularies for one fact.** Same profile, side by side: `./chromium-pipewire-camera.sh status` says `other`; `camera-relay doctor` says `DISABLED`. `other` is the least useful word available for the state a user most needs to see.

3. **`doctor`'s lists are narrower than the script's.** `doctor` iterates 5 profiles, `chromium_profile_dirs` writes 10 (adds Chrome Beta/Unstable and the three flatpaks). And the whole browser block is gated on `chromium_family`, derived from `command -v` over `firefox chromium chromium-browser brave-browser google-chrome` — so a Vivaldi-only, Edge-only or flatpak-only machine gets none of the new diagnostics even though the installer edited its profile. Once the script is on `PATH` (above), `doctor` could just shell out to `… status` and the duplicate list goes away.

4. Nits: `local … info …` at `camera-relay:1420` shadows the script's own logging function (harmless in bash, but a trap for the next editor); `Local State.camera-relay.bak` is never cleaned up on `disable`; and uninstalling with a browser open leaves the flag set — the warning prints, it just deserves a line saying what to do about it.

5. Tiny: today's upstream second branch has no `cap.capabilities & V4L2_CAP_DEVICE_CAPS` guard, but the snippet quoted in four places (and the evaluator in `doctor`) adds one. Behaviourally identical for real devices, just not verbatim.

---

## Question worth an experiment later

Your stated reason for not using `exclusive_caps=1` — which would fix Chrome, Edge **and** every Electron app at once with no profile edits — is that WirePlumber classifies the node as an output at boot, before the relay attaches.

But that's a *probe-ordering* problem, and `nudge_wireplumber` is precisely a probe-ordering fix. It re-probes after the monitor holds the writer fd, at which point an `exclusive_caps=1` node reports capture-only. So the second half of this PR arguably dissolves the objection the first half is built on.

Not asking you to change it here — there's a separate historical reason to be wary (v4l2loopback 0.12.7 couldn't switch caps via `write()`, which is why we moved to `exclusive_caps=0`), and what you shipped is verified working. But it's the only route that would help Edge and Electron apps, which this leaves broken.

Full review notes in `docs/triage/pr-83-review.md`.

PR #86, opened 2026-08-05
camera-relay: #83 review follow-ups, two browser-discovery bugs, and an Edge correction

4nrry:
Follow-up to #83, in three commits. Read in order: the review notes, then two bugs a tester found with real installs, then a correction to something #83 shipped that was wrong.

---

# 1. The non-blocking review notes — `053d02a`

Four of the five, plus the test gap the review pointed at. The fifth is answered rather than changed. Item 4's uninstall note was already handled by #84.

Thanks for checking each diagnosis independently — and for the datum that the device reports `1920x1080` discrete **while the relay is idle**. That's the load-bearing assumption under the bounded wait, and I hadn't stated it explicitly.

**1. Reversing an explicit "Disabled" is now stated, not silent.** The sharpest part is *why* a profile might hold `@2`: quite possibly this project's own earlier "leave it OFF" advice, correct for libcamera 0.2.0 and wrong here.

```
  ✓ Brave: PipeWire camera flag enabled
    (it was explicitly set to Disabled — that preference has been
     reversed; the original is saved as 'Local State.camera-relay.bak')
```

You also spotted that `enable_case "explicitly disabled"` locked in the silence — it asserted only the resulting flag list. It now has a reporting assertion beside it, plus the negative case so the wording can't leak onto profiles that simply lacked the flag.

**2. One vocabulary.** `other` is gone; `status` prints doctor's `ENABLED` / `DISABLED` / `not set`. A bare suffixless entry reads as `ENABLED`, matching what doctor already did.

**3. doctor delegates instead of keeping a second list.** The gate no longer derives "is this a Chromium machine?" from `command -v` over five binaries — which, as part 2 shows, missed exactly the machines that needed it. A profile on disk is the better signal. The inline list stays as a fallback when the tool isn't installed, and a test pins the delegation, since the failure mode is silence.

**4. Nits.** `local … info …` shadowed the script's own `info()` logging function. The backup is removed on a **successful** `disable`, but kept for profiles the running-browser guard skipped.

**5. Not changed — the `V4L2_CAP_DEVICE_CAPS` guard is upstream.** Checked rather than applied; it's line 187 of the current source:

```
187:         (cap.capabilities & V4L2_CAP_DEVICE_CAPS &&
188:          cap.device_caps & V4L2_CAP_VIDEO_CAPTURE &&
```

The only divergence in what we quote is `fd.get()` → `fd`. Recorded in `docs/triage/pr-83-review.md` with the fetched source, amend-don't-rewrite style.

---

# 2. Two bugs found by real installs — `caa96a8`

A tester installed **Vivaldi as a snap** and **Opera GX as a flatpak** and asked whether it would work out of the box on both. It would not have, for two independent reasons.

## The profile list was full paths, so it had holes

It listed native Vivaldi but **not** the snap, and **no Opera in any form**. On that machine the installer would have silently configured neither, and `doctor` would have reported nothing about them — while cheerfully claiming success.

Spelling every browser out once per packaging means three entries per browser, and nobody adds all three. The list had already drifted that way.

Replaced with config-directory *names* matched against three roots — `~/.config`, `~/snap/*/current/.config`, `~/.var/app/*/config`. A browser is named once and found in every packaging:

```
  Chrome:              ENABLED
  Vivaldi (snap):      ENABLED
  Opera GX (flatpak):  ENABLED
```

Labels carry the packaging, because two Vivaldis are otherwise indistinguishable in doctor's output and one of them may be the broken one. Paths are resolved before reporting, so a snap's `current` symlink can't list one profile twice.

## The running-browser guard failed open on flatpak

This is the one that matters. A flatpak browser records its **sandbox** PID in `SingletonLock` — for a freshly started app, a low number like 2, which on the host is a root-owned kernel thread. `kill -0` on it fails with `EPERM`, which is indistinguishable from "no such process", so the guard concluded *not running* and wrote into a live profile.

Verified against the real Opera GX flatpak: it wrote, reported `✓ flag enabled`, and the browser discarded it on exit — confirmed by the next run (after quitting) coming back as `enabled` rather than `already correct`. Exactly the silent no-op the guard exists to prevent, and worse than refusing.

Flatpak profiles now ask `flatpak ps`. Native and snap keep the PID check, which is correct there — snap doesn't use a PID namespace.

## Validated end to end across three packagings

Camera confirmed working in Chrome (deb), Vivaldi (snap) and Opera GX (flatpak) — real picture in each, and the relay journal shows three separate `Client connected` / `All clients disconnected` cycles. `doctor` reports `REAL PICTURE`, `PipeWire format: YUYV 1920x1080 — matches the device`, and still correctly reports that **Chromium cannot see the V4L2 node**, which is the whole reason the PipeWire route exists.

---

# 3. Correcting what #83 shipped about Edge — `cbc9708`

#83 put **"Edge has no such flag and cannot be fixed this way"** into both READMEs, both installers' closing hints and `doctor`. Half of that is wrong, and the half that is right was right for the wrong reason. Prompted by a reader pointing out that Edge isn't Electron — which is correct, and pulling on it showed the two cases behave oppositely.

**Edge is fixable.** `edge://flags` doesn't register the entry, so there is genuinely nothing for the installer to write into its profile — that part stands, and the exclusion from the profile list is unchanged. But the underlying Chromium feature is compiled in, so `--enable-features=WebRtcPipeWireCamera` on the command line works. Edge just can't be fixed *by editing Local State*. The compatibility tables now say **"Works, but not automatically"** and give the switch.

**Electron is not.** Grouping it with Edge understated the problem. Electron apps hit the identical V4L2 filter, but the switch does **not** help: PipeWire *camera* support isn't wired into Electron at all — only the screen-share capturer (`WebRTCPipeWireCapturer`) is. [electron#45058](https://github.com/electron/electron/issues/45058), the request to add it, is closed with no implementation. Slack, Discord, Teams and VS Code have no route today, and the docs no longer imply one.

No behaviour change — only the stated reason for an exclusion that was already correct.

## Which also corrects my own claim about `exclusive_caps=1`

I said in #83 that it was the only route that would help "Edge and Electron". It is **not** the only route for Edge; the command-line switch already is. It remains the only known route for **Electron**, since that is the direct-V4L2 path, which Electron does have.

And a third route neither of us raised: `exclusive_caps` is an `array of bool`, so `devices=2 exclusive_caps=0,1` — one node for PipeWire, one capture-only for direct V4L2 consumers — is valid syntax and would sidestep the WirePlumber regression entirely. But [v4l2loopback#134](https://github.com/v4l2loopback/v4l2loopback/issues/134) reports exactly that shape failing: with multiple devices the caps came out inconsistent (second node keeping `VIDEO_OUTPUT`, `0x85208003`) and Chromium listed only the first. Closed without resolution, so it needs testing on 0.15.x before anyone relies on it.

My position on taking `exclusive_caps=1` here is unchanged: worth its own PR with a version gate and a matrix test, not this one. It changes device setup for every user, the v4l2loopback 0.12.7 write()-can't-switch-caps history is a real hazard on older modules, and the node vanishes from every app the moment the monitor's writer fd drops. Happy to take it.

---

## Verification

- **143 assertions across 7 suites, all passing** (59 in this suite, up from 41).
- `shellcheck -S warning` clean, no new warnings versus `main`.
- Both doctor paths exercised on hardware — delegated and fallback.
- Camera validated in three browsers across three packagings: deb, snap, flatpak.


Andycodeman (approved):
Merging. Verified on current `main` (which now carries #84 and #87): merges cleanly, **163 assertions across 8 suites all pass** on the merged result, `bash -n` clean. You handled the `flag_tool` rename from #84 correctly.

## You're right and I was wrong about the DEVICE_CAPS guard

Thanks for checking rather than applying it. I fetched and decoded the source myself to be sure, and line 187 is exactly what you say:

```
187:         (cap.capabilities & V4L2_CAP_DEVICE_CAPS &&
188:          cap.device_caps & V4L2_CAP_VIDEO_CAPTURE &&
```

My review took a summarised fetch at face value and reported the summary as the source. The correction in `docs/triage/pr-83-review.md` is the right way to record it — amend, don't rewrite.

Your fetch also surfaces a third gate I never mentioned: `HasUsableFormats(fd.get(), cap.capabilities)` on line 190. Worth knowing that a node can pass the caps test and *still* be dropped for having no usable format — another way to arrive at "Chrome sees nothing".

## The flatpak guard bug is the best find here

A sandboxed browser records its **sandbox** PID in `SingletonLock`; PID 2 on the host is `kthreadd`, root-owned; `kill -0` returns EPERM, which is indistinguishable from ESRCH — so the guard concluded "not running" and wrote into a live profile. Silent no-op, the one outcome worse than refusing, and it only shows up with a real flatpak install. Good catch, and the `flatpak ps` fallback ("no CLI to ask → treat a present lock as running") errs in the correct direction.

The config-directory-names refactor is the right call too. Three entries per browser was always going to drift, and it demonstrably had.

## Verified on hardware

Same Book4 Ultra. New discovery finds the same two profiles with the corrected vocabulary, and the glob scan correctly ignores about ten non-browser snap and flatpak app directories — no false positives:

```
  Chrome:              not set        (was: absent)
  Brave:               DISABLED       (was: other)
```

Doctor's delegation works end to end, including the remedy still printing for a non-`ENABLED` delegated profile.

---

## One thing I'm fixing in a follow-up

**The feature name is misspelled in two places, including one this PR adds.** `--enable-features` silently ignores unknown names and Chromium feature names are case-sensitive. From the source — `media/capture/capture_switches.cc:49`:

```c
BASE_FEATURE(kWebRtcPipeWireCamera, base::FEATURE_DISABLED_BY_DEFAULT);
```

So the string is `WebRtcPipeWireCamera`. Two places still say `WebRTCPipeWireCamera`, which does nothing:

- `webcam-fix-libcamera/README.md:176` — inherited from #83, and worse than a typo: that line tells Electron users to launch with the switch, while this PR adds text further down saying the switch **does not help Electron**. The README ends up contradicting itself *and* naming a feature Chromium ignores.
- `camera-relay/chromium-pipewire-camera.sh:119` — the new `Deliberately absent:` comment.

Everything else in the PR uses the correct spelling, so this is two lines plus rewriting that stale README paragraph.

## Non-blocking

1. **`flatpak ps --columns=application | grep -qxF "$appid"` is the SIGPIPE-under-pipefail pattern you documented above `loopback_real_format`.** The script runs `set -uo pipefail`, and I confirmed `seq … | grep -qxF` returns 141 there. `flatpak ps` output is small enough that the writer always finishes first, so it does not bite today — but if it ever did, the result is `return 1` = "not running" = write into a live profile, which is precisely the direction this function exists to prevent. Buffering first costs nothing; I'll fold it into the same follow-up.

2. **The delegated lines lost their label.** doctor now prints `Chrome:              not set` where it printed `Chrome PipeWire flag: not set`. Three lines above, `google-chrome: /usr/bin/google-chrome` uses the same `name:` shape to mean a binary path — two different meanings in one visual form, and the flag lines are the ones people are scanning for.

3. `[[ "$flag_status" == *"No Chromium-family"* ]]` couples doctor to an exact sentence in another script's output. An exit code or a `--quiet` mode would survive rewording.

4. The `else` branch keeps the fallback `for prof` loop at its old indentation, so it reads as though it were still unconditional.

5. Body says "143 assertions across 7 suites" — against current `main` it is 163 across 8. `test-unit-regeneration.sh` landed in #87 while this was open.

## On `exclusive_caps`

Agreed on all three points, and the `devices=2 exclusive_caps=0,1` idea plus [v4l2loopback#134](https://github.com/v4l2loopback/v4l2loopback/issues/134) is a genuinely useful lead I did not have — a per-node split would sidestep the WirePlumber regression entirely if 0.15.x behaves. Own PR, version gate, matrix test. Yours if you want it.

Andycodeman:
Merged and shipped as [v0.3.64](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/releases/tag/v0.3.64), with the feature-name fix landed on top in #88.

Worth flagging what that turned out to be, since it cuts across your Edge/Electron correction: `--enable-features` **silently ignores a name it does not recognise**, and the names are case-sensitive. From `media/capture/capture_switches.cc:49`:

```c
BASE_FEATURE(kWebRtcPipeWireCamera, base::FEATURE_DISABLED_BY_DEFAULT);
```

So `WebRTCPipeWireCamera` — the spelling in `webcam-fix-libcamera/README.md:176` from #83, and in this PR's new `Deliberately absent:` comment — was doing nothing. Every other site in your PR had it right.

That README line was the awkward one: it told Electron users to launch with the switch, while this PR added text further down establishing that the switch does not help Electron. Rewritten to point at your detailed section. Your comment also still carried the pre-#86 framing of Edge ("ships no enable-webrtc-pipewire-camera flag") — updated to match what the rest of the PR establishes.

There is now a test asserting every `--enable-features=` use site spells it correctly, matching only use sites so a comment naming the wrong spelling to warn about it does not trip.

I also folded in the `flatpak ps` buffering. Confirmed the pattern is live in this shell:

```
$ bash -c 'set -uo pipefail; seq 1 500000 | grep -qxF 3; echo rc=$?'
rc=141
```

Never fires for `flatpak ps` output sizes, but it would have landed on `return 1` = "not running", which is the direction your whole fix is about.

The other three notes (delegated lines losing their `PipeWire flag:` label, the `No Chromium-family` string coupling, the fallback loop's indentation) I left alone — none change behaviour, and they are yours to take or drop.

Standing offer on `exclusive_caps`: the `devices=2 exclusive_caps=0,1` lead with v4l2loopback#134 is the most interesting thread here. Yours if you want it.

PR #90, opened 2026-08-09
webcam-fix-libcamera: warn against the intel-ipu6-dkms offer in Additional Drivers

4nrry:
## What

Two new troubleshooting sections in `webcam-fix-libcamera/README.md`, plus a cross-link from the existing `v4l2-relayd` section.

- **`### "Additional Drivers" offers intel-ipu6-dkms — don't install it`** — preventive note: why the panel is wrong, and what installing it costs.
- **`### Camera stopped working after installing intel-ipu6-dkms`** — recovery runbook for anyone who already accepted the offer.

Docs only. No code, no installer changes.

## Why

Ubuntu's Additional Drivers panel (`software-properties-gtk`) lists the IPU as **"This device is not working"** and offers `intel-ipu6-dkms` — on a machine where the camera is already streaming fine through the in-tree driver:

> Intel Corporation: Meteor Lake IPU
> This device is not working.
> ( ) Using Intel Integrated Image Processing Unit 6 (IPU6) driver from intel-ipu6-dkms (open source)
> (•) Do not use the device

The label is a **false positive**. `ubuntu-drivers` matches by modalias, and the only modalias `intel-ipu6-dkms` declares is for `intel-ipu6-psys` — the PSYS function, which the mainline driver does not expose. Nothing claims it, so the device reads as unclaimed even while `intel-ipu6` is bound and `Connected 1 cameras`.

Accepting the offer can **break the camera**. `intel-ipu6-dkms` is the out-of-tree [`intel/ipu6-drivers`](https://github.com/intel/ipu6-drivers) stack, and its `dkms.conf` builds **`ov02c10` unconditionally** into `/updates` — the same module name and the same directory as our `ov02c10-26mhz-fix`. Intel's copy has no 26 MHz patch, so an affected board goes straight back to `external clock 26000000 is not supported`.

And there is nothing to gain in exchange:

| Module | Built on kernel ≥ 6.10? | Useful here? |
|---|---|---|
| `ov02c10` | yes, unconditional | **no** — overwrites our patched copy |
| `intel-ipu6`, `intel-ipu6-isys` | no (gated on < 6.10) | in-tree drivers already own this |
| `intel-ipu6-psys` | yes, unconditional | only serves the proprietary HAL |
| `cio2-bridge` | not in `dkms.conf` | — (`ipu-bridge-fix` is unaffected) |

`libcamhal0`, `intel-ipu6-camera-hal` and `gstreamer1.0-icamera` have **no installation candidate** on Ubuntu 26.04, so the PSYS module has nothing to feed.

The repo had zero mentions of `intel-ipu6-dkms`, `ubuntu-drivers` or "Additional Drivers" anywhere in the tree — so every Ubuntu user meets this offer with no warning, and the obvious move is to accept it.

## Reviewer notes

- **Placement** is next to the 26 MHz entry rather than next to the `v4l2-relayd` one, because the collision is specifically on `ov02c10`. A reader who lands on the clock error needs to know the Intel DKMS may be the cause. The `v4l2-relayd` section gets a one-paragraph pointer instead — `intel-ipu6-dkms` is the kernel-side half of the same OEM stack.
- **Secure Boot** is called out in the runbook: an unsigned DKMS module is rejected silently and the kernel falls back to the in-tree driver, which rejects 26 MHz too. Links to the existing `ov02c10-26mhz-fix/README.md#secure-boot` rather than restating it.
- **The `lspci` line uses device IDs** (`-d 8086:7d19`, with `8086:a75d` in a comment) instead of grepping the description — lspci classifies the IPU as *Multimedia controller*, not *Image Processing*, so the obvious grep finds nothing.

## Verification

Every command in the new sections was run on an affected board — Galaxy Book4 Ultra `NP960XGL-XG1BR` (`960XGL`), Meteor Lake IPU6 `8086:7d19`, Ubuntu 26.04, kernel `7.0.0-29-generic`, **Secure Boot enabled**:

```
$ lspci -nnk -d 8086:7d19
00:05.0 Multimedia controller [0480]: Intel Corporation Meteor Lake IPU [8086:7d19] (rev 04)
        Kernel driver in use: intel-ipu6

$ modinfo -n ov02c10
/lib/modules/7.0.0-29-generic/updates/dkms/ov02c10.ko.zst

$ journalctl -k -b -g '26000000Hz clock'
ov02c10 i2c-OVTI02C1:00: 26000000Hz clock (not 19200000Hz): sensor runs fast; ... ~30fps
```

Claims about the package were checked against the archive, not from memory: `apt-cache show intel-ipu6-dkms` for the modalias, the `.deb` file listing on packages.ubuntu.com for the shipped sources, and the upstream `dkms.conf` for the build conditionals and `DEST_MODULE_LOCATION`.

Markdown checked: heading order, balanced fences, and the in-page anchor slug (the heading carries quotes, backticks, an em dash and an apostrophe, so the anchor needs the double hyphen).

The recovery commands are documented, not executed — nothing on the test machine was changed.

### DIFF
--- a/webcam-fix-libcamera/README.md
+++ b/webcam-fix-libcamera/README.md
@@ -286,6 +286,76 @@ Some Galaxy Book3/Book4 models have a 26 MHz external clock instead of the expec
 cd ov02c10-26mhz-fix && sudo ./install.sh
 ```
 
+### "Additional Drivers" offers `intel-ipu6-dkms` — don't install it
+
+Ubuntu's **Additional Drivers** panel (`software-properties-gtk`) lists the IPU
+as *"Intel Corporation: Meteor Lake IPU — This device is not working"* and
+offers *"Using Intel Integrated Image Processing Unit 6 (IPU6) driver from
+intel-ipu6-dkms (open source)"*. Leave it on **"Do not use the device"**.
+
+The "not working" label is a false positive. `ubuntu-drivers` matches hardware
+by modalias, and the only modalias `intel-ipu6-dkms` declares is for
+`intel-ipu6-psys` — the PSYS function, which the mainline driver does not
+expose. No installed package claims that modalias, so the panel reports the
+device as unclaimed even while the in-tree driver is bound and streaming:
+
+```bash
+lspci -nnk -d 8086:7d19    # Meteor Lake; use 8086:a75d on Raptor Lake
+```
+
+That prints `Kernel driver in use: intel-ipu6` on a working system.
+
+Installing the offered driver can break the camera. `intel-ipu6-dkms` is the
+out-of-tree Intel stack ([`intel/ipu6-drivers`](https://github.com/intel/ipu6-drivers),
+what the legacy [`webcam-fix/`](../webcam-fix/) is built around), and its
+`dkms.conf` builds **`ov02c10` unconditionally** into `/updates` — the same
+module name and the same directory as the [26 MHz fix](../ov02c10-26mhz-fix/).
+Intel's copy does not carry the 26 MHz patch, so on an affected board the camera
+goes back to `external clock 26000000 is not supported`.
+
+There is nothing to gain in exchange. On any kernel 6.10 or newer, `intel-ipu6`
+and `intel-ipu6-isys` are not built at all (the `dkms.conf` gates them on older
+kernels, because the in-tree drivers took over). What remains is
+`intel-ipu6-psys`, which is only useful to the proprietary camera HAL —
+and `libcamhal0`, `intel-ipu6-camera-hal` and `gstreamer1.0-icamera` have no
+installation candidate on Ubuntu 26.04.
+
+### Camera stopped working after installing `intel-ipu6-dkms`
+
+**Symptom.** The camera worked, you accepted the Additional Drivers offer above,
+and now `camera-relay status` fails or dmesg is back to
+`external clock 26000000 is not supported`.
+
+Check which `ov02c10` the kernel resolves, and who owns it:
+
+```bash
+modinfo -n ov02c10
+dkms status | grep -E 'ov02c10|ipu6'
+```
+
+Remove the Intel stack and reinstate the patched driver:
+
+```bash
+sudo apt purge intel-ipu6-dkms
+sudo dkms install ov02c10/1.0 -k "$(uname -r)"
+sudo depmod -a
+sudo update-initramfs -u
+```
+
+Reboot, then verify. The path must land under `updates/dkms`, not
+`kernel/drivers`, and the clock line must be the patched driver's:
+
+```bash
+modinfo -n ov02c10
+journalctl -k -b -g '26000000Hz clock'
+```
+
+If `modinfo` still points at `kernel/drivers` after the reboot, the DKMS build
+either failed or was rejected. With Secure Boot enabled an unsigned module is
+rejected in silence and the kernel falls back to the in-tree driver — which
+rejects 26 MHz too. See
+[Secure Boot](../ov02c10-26mhz-fix/README.md#secure-boot) in the 26 MHz fix.
+
 ### Too many "ipu6" entries in camera list
 
 Log out and back in for the udev rules and WirePlumber config to take effect, then check with:
@@ -491,6 +561,11 @@ disables and masks `v4l2-relayd.service` and moves the OEM
 `/etc/modprobe.d/v4l2loopback.conf` aside (restored by `uninstall.sh`). Just
 re-run `sudo bash install.sh` and reboot. The change is fully reversible:
 `uninstall.sh` restores the OEM file and re-enables the service.
+
+The kernel-side half of the same OEM stack is `intel-ipu6-dkms`, which Ubuntu's
+Additional Drivers panel offers on these machines — see
+["Additional Drivers" offers `intel-ipu6-dkms`](#additional-drivers-offers-intel-ipu6-dkms--dont-install-it)
+above. Don't install it.
 
 ### LED on but black image, on laptops with a dedicated GPU
 
--- a/webcam-fix-libcamera/README.md
+++ b/webcam-fix-libcamera/README.md
@@ -316,9 +316,13 @@ goes back to `external clock 26000000 is not supported`.
 There is nothing to gain in exchange. On any kernel 6.10 or newer, `intel-ipu6`
 and `intel-ipu6-isys` are not built at all (the `dkms.conf` gates them on older
 kernels, because the in-tree drivers took over). What remains is
-`intel-ipu6-psys`, which is only useful to the proprietary camera HAL —
-and `libcamhal0`, `intel-ipu6-camera-hal` and `gstreamer1.0-icamera` have no
-installation candidate on Ubuntu 26.04.
+`intel-ipu6-psys`, which is only useful to the proprietary camera HAL — the
+[legacy stack](../webcam-fix/) this fix replaces. That HAL is not in the Ubuntu
+archive on any release: `libcamhal-ipu6epmtl` and `gstreamer1.0-icamera` are
+published only in the `ppa:oem-solutions-group/intel-ipu6` OEM PPA, which is not
+enabled on a stock install. So out of the box `intel-ipu6-psys` has nothing to
+feed — and if you *do* enable that PPA, you are setting up the legacy stack
+deliberately, which is a different decision from repairing this one.
 
 ### Camera stopped working after installing `intel-ipu6-dkms`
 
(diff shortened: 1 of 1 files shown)


Andycodeman (changes requested):
Thank you again, @4nrry — this is a really valuable catch. The repo had zero mentions of `intel-ipu6-dkms`, `ubuntu-drivers` or "Additional Drivers" anywhere, so every Ubuntu user has been meeting that offer with no warning at all, on a machine where the panel is actively telling them their working camera is broken. That's exactly the kind of trap worth documenting.

I checked the diagnosis rather than taking it on trust, and **all of it holds**:

* `apt-cache show intel-ipu6-dkms` declares exactly one modalias entry, `intel-ipu6-psys`, and it covers both `d00007D19` (Meteor Lake) and `d0000A75D` (Raptor Lake). Your explanation of the false positive is right, and it applies to both of our platforms.
* Upstream `intel/ipu6-drivers` `dkms.conf` confirms the table: `BUILT_MODULE_NAME[2]="ov02c10"` with `DEST_MODULE_LOCATION[2]="/updates"`, unconditional; `intel-ipu6` and `intel-ipu6-isys` gated behind `version_lt ${KERNEL_VERSION} 6.10.0`; `intel-ipu6-psys` unconditional; no `cio2-bridge`. Same module name and same destination as `ov02c10-26mhz-fix/dkms.conf`, so the collision is real.
* Both dmesg strings match the driver source — `external clock 26000000 is not supported` is the `dev_err_probe` rejection path, and `26000000Hz clock` is the patched driver's own `dev_info`. Good choice for a "did the patched driver actually load" check.
* The in-page anchor is correct, double hyphen and all; the three relative links resolve; the `lspci -nnk -d 8086:7d19` line works as written; placement next to the 26 MHz entry is the right call for the reason you gave.

The preventive section I'd merge as-is. It's the **recovery runbook** that needs work before this goes in — there's one command in it that doesn't do anything.

---

## 🔴 Blocking — `dkms install ov02c10/1.0` exits without reinstating anything

```bash
sudo apt purge intel-ipu6-dkms
sudo dkms install ov02c10/1.0 -k "$(uname -r)"
```

Purging `intel-ipu6-dkms` clears *Intel's* DKMS state. It doesn't touch ours — the `ov02c10/1.0` "installed" marker for the running kernel survives, and DKMS checks that marker before doing anything else:

```sh
# /usr/sbin/dkms:1244  (dkms 3.0.11)
is_module_installed "$module" "$module_version" "$kernelver" "$arch" && die 5 \
    $"This module/version combo is already installed for kernel $kernelver ($arch)."
```

`_is_module_installed` (`:1432`) is purely a symlink test on `$dkms_tree/ov02c10/kernel-$KVER-$arch`, which the Intel purge never clears. So the command dies with exit 5, the `depmod -a` and `update-initramfs -u` that follow run over an unchanged tree, and the user reboots into exactly the same broken camera the section is meant to fix.

Two smaller things in the same block:

* It skips Secure Boot signing entirely. `ov02c10-26mhz-fix/install.sh` (L95–163) is what writes `/etc/dkms/framework.conf.d/ov02c10-mok-keys.conf`, verifies the module came out signed, and queues MOK enrollment. A bare `dkms install` gets none of that — and on your own board, with Secure Boot on, a rebuild without it would be rejected at load.
* `depmod -a` + `update-initramfs -u` are a partial reimplementation of `install.sh` L248–258, which already does both, with `dracut`/`mkinitcpio` fallbacks for non-Ubuntu.

**Suggested replacement** for "Remove the Intel stack…" and its code block:

> Remove the Intel stack, then re-run the 26 MHz installer to rebuild and
> reinstate the patched driver:
>
> ```bash
> sudo apt purge intel-ipu6-dkms
> cd ov02c10-26mhz-fix && sudo ./install.sh
> ```
>
> Don't reach for `dkms install ov02c10/1.0` directly — purging the Intel
> package does not clear DKMS's own "installed" marker for `ov02c10/1.0`, so
> `dkms install` exits with *"This module/version combo is already installed
> for kernel …"* and changes nothing. `install.sh` does the full
> `dkms remove --all` → `add` → `build` → `install` cycle, re-signs the module
> for Secure Boot, and refreshes `depmod` and the initramfs.

That also lines up with how this README already handles the *other* OEM-stack collision a few sections down ("Just re-run `sudo bash install.sh` and reboot"), and with the invocation form used in the 26 MHz entry right above.

## 🟠 Should fix — the `modinfo` check can't detect the failure it's used to diagnose

> If `modinfo` still points at `kernel/drivers` after the reboot, the DKMS build either failed **or was rejected**. With Secure Boot enabled an unsigned module is rejected in silence…

`modinfo -n` resolves the path out of `modules.dep`, which prefers `updates/dkms` regardless of whether the module can actually be loaded. Signature rejection happens at *load* time and never changes what `modinfo` prints. So the Secure Boot case — the one the paragraph spends its remaining lines on — shows the "good" path, passes the test, and leaves the reader with a dead camera and no next step.

Worth splitting into two checks. `install.sh` L212 already uses the right one:

> ```bash
> modinfo -n ov02c10                 # must be under updates/dkms
> modinfo ov02c10 | grep -i '^sig'   # no output = unsigned
> journalctl -k -b -g '26000000Hz clock'
> ```
>
> A path under `kernel/drivers` means the DKMS build or install failed. The
> right path with no `sig*` fields means the module was built but is unsigned,
> so under Secure Boot it is rejected in silence and the 26 MHz error persists.
> See [Secure Boot](../ov02c10-26mhz-fix/README.md#secure-boot) in the 26 MHz fix.

## Nits (take or leave)

1. **"There is nothing to gain in exchange" is only true on 26.04.** The sentence itself is scoped correctly — the `oem-solutions-group/intel-ipu6` PPA is noble-only, so those packages genuinely have no candidate on 26.04 — but the paragraph's conclusion reads as universal. On my 24.04 box with that PPA enabled, `libcamhal0` and `gstreamer1.0-icamera` are both installed and available. The argument still holds for this README's audience, since the libcamera path supersedes the HAL path, but a clause noting that would stop a 24.04 reader from spotting the gap and discounting the whole warning.
2. **Package naming.** The repo elsewhere calls the HAL `libcamhal-ipu6epmtl` (`webcam-fix/README.md` L229, L233). `intel-ipu6-camera-hal` isn't a package name apt knows about on noble at all. Worth aligning with the names already used in-tree.

---

Fix the blocker and I'll merge this. The `modinfo` one I'd like too, since a reader who hits it is by definition already in the failure case — but if you'd rather land the runbook fix on its own, I'll take that and file the rest.

4nrry:
Thanks — the blocker was real, and I'd shipped a runbook that does nothing. Pushed in 9262544.

**🔴 `dkms install` exits 5.** Confirmed on my box before changing anything. dkms here is 3.2.2 rather than your 3.0.11, so the line numbers moved (the `die` is at `/usr/sbin/dkms:1662`, `_is_module_installed()` at `:1840`), but the behaviour is identical — and the marker you predicted is sitting right there:

```
$ ls -l /var/lib/dkms/ov02c10/kernel-7.0.0-29-generic-x86_64
... -> 1.0/7.0.0-29-generic/x86_64
```

Purging the Intel package never touches it. Replaced the block with your suggested `cd ov02c10-26mhz-fix && sudo ./install.sh`, plus the paragraph explaining why not to reach for `dkms install` — otherwise the next person rediscovers this the same way.

**🟠 `modinfo -n` can't see signature rejection.** Also right, and this was the worse of the two: the paragraph pointed at Secure Boot while using a check that passes in exactly that case. Took your split verbatim and added one clause naming the mechanism (`modinfo -n` reads `modules.dep`; rejection is at load time), so the reason the second check exists is on the page.

**Nit 2, package naming.** Fixed — `libcamhal-ipu6epmtl` and `gstreamer1.0-icamera`, matching `webcam-fix/README.md` L233–234. `intel-ipu6-camera-hal` was mine, from the non-apt distro names in the uninstaller.

**Nit 1 — I checked this one and your premise is off, though your conclusion stands.** The PPA is not noble-only. Launchpad publishes `ipu6-camera-hal` for focal, groovy, hirsute, jammy, noble, oracular **and resolute**, and the 26.04 build is currently `Published`:

```
$ ... getPublishedSources source_name=ipu6-camera-hal   # resolute only
  Published    0~git202601200757.9899efa~ubuntu26.04.2
  Superseded   0~git202601200757.9899efa~ubuntu26.04.1
  Deleted      0~git202506270918.c933525-1~ubuntu26.04.1
```

So "no candidate on 26.04" was true on my machine for a duller reason than either of us gave: the PPA simply isn't enabled here. Had I scoped the sentence by release as planned, it would have been a fresh wrong claim.

Rewrote it release-independently instead: the HAL is absent from the Ubuntu archive on *every* release and ships only via `ppa:oem-solutions-group/intel-ipu6`, which isn't enabled on a stock install — so out of the box `intel-ipu6-psys` has nothing to feed, and enabling that PPA is a deliberate move to the legacy stack, not a repair of this one. That holds on your 24.04 box too, which I think was your actual point.

Diff on this push: +27/−14, still the one file.

Andycodeman (approved):
Both findings fixed, and thank you for the pushback on nit 1 — you were right and I wasn't.

**🔴 Blocker.** Runbook now purges then re-runs `ov02c10-26mhz-fix/install.sh`, and you kept the paragraph explaining *why* `dkms install ov02c10/1.0` isn't the repair command. That was the right call — swapping the command silently would have left the next person to rediscover it exactly as we did. Link resolves, file is executable. And you reproduced the guard on dkms **3.2.2** (`die` at `:1662`, `_is_module_installed()` at `:1840`) with the surviving marker symlink, rather than taking my 3.0.11 line numbers on trust — that's now confirmed on two dkms versions instead of one.

**🟠 The `modinfo` split.** Better than what I suggested. Naming the mechanism on the page — `modinfo -n` reads `modules.dep`, rejection happens at load — means the second check reads as necessary rather than as belt-and-braces, so nobody trims it later.

**Nit 1.** My reason was wrong, and wrong in a way I should have caught: I inferred "the PPA is noble-only" from the filename of the sources entry on my box (`…-intel-ipu6-noble.sources`), which only records which series *this machine* subscribes to — it says nothing about what the PPA publishes. Launchpad has `ipu6-camera-hal` for focal through resolute, with 26.04 currently `Published` at `0~git202601200757.9899efa~ubuntu26.04.2`, exactly as you found. Had you scoped the sentence by release as I asked, I'd have talked you into a fresh wrong claim. The release-independent rewrite is the better fix, and the archive-absence framing is the durable one — I confirmed here that `libcamhal-ipu6epmtl` and `gstreamer1.0-icamera` resolve only to the PPA with no archive entry at all.

**Nit 2** fixed too.

One residual I'm not holding this for, just so it's on the record: `grep -i '^sig'` catches *unsigned*, but a module signed with a MOK key that was never enrolled still shows `sig*` fields and is still rejected. The Secure Boot link covers enrollment and `install.sh` queues it and warns when it's pending, so a reader lands in the right place either way.

Merging. Thanks again — this one started as a documentation gap and ended up correcting a real trap in the recovery path, plus a wrong belief of mine about the PPA. Genuinely good work.

PR #91, opened 2026-08-20
fix: webcam-fix-libcamera detects Fedora as Arch when pacman is installed

dpol1:
## The failure

On Fedora 43, `webcam-fix-libcamera/install.sh` printed:

```
✓ Arch-based detected
```

and later died at `[9/14] Installing libcamera...` trying to install Arch package names with pacman:

```
error: target not found: git
error: target not found: meson
error: target not found: ninja
error: target not found: gcc
```

Root cause: the script identifies the distro by whichever package-manager binary exists, checking `pacman` **before** `dnf`. But a package manager binary does not uniquely identify the OS — Fedora ships a real `pacman` package (`pacman-7.0.0-5.fc43.x86_64` here, installable from the standard repos), so any Fedora host that has it is misdetected as Arch.

## The fix

Detection now reads `/etc/os-release` first — `ID`, then the `ID_LIKE` ancestry list — and only falls back to probing for package managers when os-release gives no answer (the pre-existing behavior, kept as a last resort for unusual systems). The logic moves into a `detect_distro()` function so it can be tested in isolation.

Mapping is unchanged for everything that worked before:

- `ID=arch` or `ID_LIKE` containing `arch` (EndeavourOS, Manjaro, …) → `arch`
- `ID=fedora` or `ID_LIKE` containing `fedora` (Nobara, RHEL family) → `fedora`
- `ID=ubuntu|pop|linuxmint` or `ID_LIKE` containing `ubuntu` (checked before `debian`, as before) → `ubuntu`
- `ID=debian` / `ID_LIKE` containing `debian` → `debian`
- nothing recognizable → same "Unsupported distro" error

## Tests

Added `camera-relay/tests/test-distro-detection.sh` (same conventions as the existing suites there): extracts `detect_distro()` from the script and runs it against fixture os-release files with a stub-only `PATH`, so the host system never leaks in. 16 checks: the Fedora-with-pacman regression, all supported distros and derivatives, the no-os-release fallback, rejection of unsupported systems, plus wiring checks that guard against the block reverting to a bare binary probe.

```
$ camera-relay/tests/test-distro-detection.sh
...
16 passed, 0 failed
$ camera-relay/tests/test-pipewire-restart-guard.sh   # existing suite, still green
17 passed, 0 failed
```

Also verified live on the affected Fedora 43 machine (with `/usr/bin/pacman` present): detection now reports `Fedora/DNF-based`.

## Test plan

- [x] `bash -n webcam-fix-libcamera/install.sh`
- [x] New test suite: 16/16 pass
- [x] Existing `test-pipewire-restart-guard.sh`: 17/17 pass
- [x] Live run of the detection step on Fedora 43 with pacman installed

## Note on sibling scripts

`webcam-fix/install.sh`, `webcam-fix-book5/install.sh` and `webcam-fix/uninstall.sh` share the same pacman-first probe and are misdetected the same way; `mic-fix`, `speaker-fix` and `speaker-fix-940xfg` happen to check `dnf` first and are unaffected. Left out of this PR to keep it reviewable — happy to follow up with the same treatment if you want it.

### DIFF
--- a/camera-relay/tests/test-distro-detection.sh
+++ b/camera-relay/tests/test-distro-detection.sh
@@ -1,0 +1,176 @@
+#!/usr/bin/env bash
+# Regression tests for distro detection in webcam-fix-libcamera/install.sh.
+#
+# The script used to identify the distro purely by which package manager
+# binary existed, checking `pacman` first. That heuristic is wrong on any
+# distro that ships another distro's package manager — and Fedora does: it
+# packages a real `pacman` (e.g. pacman-7.0.0-5.fc43 on Fedora 43). A Fedora
+# host with it installed was greeted with "✓ Arch-based detected" and the
+# install died at step [9/14] trying Arch package names:
+#
+#   error: target not found: git
+#   error: target not found: meson
+#
+# Detection now reads /etc/os-release (ID, then the ID_LIKE ancestry list)
+# and probes for package managers only when os-release gives no answer.
+# These tests drive the extracted detect_distro() against fixture os-release
+# files and a stub-only PATH, so they need no root and touch no system state.
+#
+# Usage: ./test-distro-detection.sh
+
+set -uo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+INSTALL="$ROOT/webcam-fix-libcamera/install.sh"
+PASS=0
+FAIL=0
+
+ok()  { echo "  ✓ $*"; PASS=$((PASS + 1)); }
+bad() { echo "  ✗ $*"; FAIL=$((FAIL + 1)); }
+
+TMP=$(mktemp -d)
+trap 'rm -rf "$TMP"' EXIT
+
+echo "test-distro-detection ($INSTALL)"
+
+# The function is pure bash (only builtins), so the subshell can run with
+# PATH containing nothing but the stub package managers — `command -v` then
+# sees exactly the set each scenario declares, never the host's binaries.
+FN_SRC=$(sed -n '/^detect_distro()/,/^}/p' "$INSTALL")
+if [[ -z "$FN_SRC" ]]; then
+    bad "detect_distro() not found in install.sh — tests cannot run"
+    echo; echo "$PASS passed, $FAIL failed"
+    exit 1
+fi
+
+# run_detect <os-release file or -> [pm...] → "DISTRO|DISTRO_LABEL" on stdout.
+# "-" means no os-release (a path that does not exist). Exits 9 on rejection.
+run_detect() {
+    local os_release="$1"; shift
+    [[ "$os_release" == "-" ]] && os_release="$TMP/does-not-exist"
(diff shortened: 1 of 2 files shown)


dpol1:
This a real gap found while using the script, the fix is agent maded. I use Fedora btw :-)

Andycodeman:
Merged as `dc6157d3` and shipped as [v0.3.67](https://github.com/Andycodeman/samsung-galaxy-book-linux-fixes/releases/tag/v0.3.67). Thank you — this is a genuinely good PR and I want to say why, because "agent made" undersells it.

You found it by actually using the thing, you diagnosed the real cause rather than the symptom, and the fix is the correct general one instead of a special case for Fedora. That's the part that matters: `command -v pacman` was never going to be a reliable way to identify an OS, and reading `ID` / `ID_LIKE` is what should have been there from the start.

A few notes from the review:

**I checked your claims rather than taking them on faith, and they hold.** `pacman-7.0.0-5.fc43.x86_64` is right there in Fedora's package metadata, summary and all — "Package manager for the Arch distribution ... can be used to install Arch into a container or to recover an Arch installation from a Fedora system." Exactly the NEVRA you cited.

**The test suite is the best part.** I mutation-tested it: hoisted the pacman probe back above the os-release mapping inside `detect_distro()`, and exactly one of the sixteen checks failed — the Fedora one — with the rest still green. That's the property that makes a regression suite worth having, and plenty of suites don't have it. The stub-only `PATH` trick is neat too, since the function is pure builtins and the host can't leak in. Full suite is 198 assertions across 10 suites now, all green on the merge commit.

**One correction to the PR description.** You wrote that `mic-fix`, `speaker-fix` and `speaker-fix-940xfg` "happen to check `dnf` first and are unaffected." They're unaffected by *this* case, but Arch's `extra` repo ships `dnf 4.24.0` and `apt 3.3.3` — so a dnf-first probe misdetects an Arch host as Fedora. Same defect, mirrored. Worth knowing before the follow-up so nobody reads "checks dnf first" as "safe."

**Yes please to the follow-up**, and the scope is bigger than the three you listed — it's `webcam-fix/install.sh`, `webcam-fix/uninstall.sh`, `webcam-fix-book5/install.sh` (pacman-first) plus those three above (dnf-first). One question worth settling before you start: six copies of `detect_distro()` is how the current situation came about, but there's no shared shell library in the repo today, so where it should live is a real decision rather than a mechanical port. Happy to take your view on it — the installers are standalone by design so people can `curl | bash` a single file, which cuts against a `lib/` directory, but I'm open.

Two non-blocking nits I didn't hold the merge for, in case you touch this code again: `while IFS='=' read -r key value < file` silently drops the last line if the file has no trailing newline, and the same for CRLF endings — and in both cases it falls through to the package-manager probe, i.e. straight back into the bug. `|| [[ -n "$key" ]]` on the `while` closes the first. Every real distro newline-terminates `os-release`, so this is theoretical; it's only worth mentioning because of where the fall-through lands.

I use Ubuntu, so I couldn't reproduce the original failure here — your live Fedora 43 confirmation is the on-hardware evidence for this one. Appreciated.

